PR #3, opened 2016-05-19
Change umbrella application name to nervy

Gazler:
This helps distinguish between umbrella the concept, and umbrella the
app name.


mobileoverlord:
Sounds good, Thanks :bowtie:

PR #9, opened 2016-05-31
Correction to fw / firmware app refs

joelbyler:
User Interface doc has instructions to create a sub-app with the name `fw` but then its referenced as `firmware` elsewhere in the doc, just cleaning this up to make it less confusing


mobileoverlord:
Thanks! :bowtie:

PR #10, opened 2016-05-31
Add note about rebar -v >=2.6.1

joelbyler:
Justin helped me thru this issue on slack so I thought I would try and add the note to the docs for others to see in case they run into the same issue.  Hopefully it will help.

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -69,7 +69,7 @@ When it's time to create your firmware:
 $ mix firmware
 ```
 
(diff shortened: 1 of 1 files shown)


mobileoverlord:
@joelbyler can you update this ask them to just run `mix local.rebar` to update, I'll still merge it since people don't update local rebar often

joelbyler:
Thanks for the suggestion @mobileoverlord, I removed references to an explicit version as well.  Hopefully this works for you.  Let me know if there is anything else.

mobileoverlord:
Thanks for the contribution! :bowtie:

PR #13, opened 2016-06-07
Added some definitions; other small edits for clarification

ChuckLutz:
Please check the new definitions I added; for "assemble", we had "The process of combining system, application, and configuration into a firmware bundle" so I wanted to include "application" and "configuration", so I took a stab at them

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -22,6 +22,8 @@ host | The computer on which you are editing source code, compiling, and assembl
 target | The platform for which your firmware is built (for example, Raspberry Pi, Raspberry Pi 2, or Beaglebone Black)
 toolchain | The tools required to build code for the target, such as compilers, linkers, binutils, and C runtime
 system | A lean buildroot-based linux distribution that has been customized and cross-compiled for a particular target
+application | The Elixir (and any native, e.g. C) source code for your project. Note that this may entail more than one "application" in Elixir parlance.
+configuration | The settings and other target-specific files that steer the firmware building process for your particular target
 assemble | The process of combining system, application, and configuration into a firmware bundle
 firmware bundle | A single file that contains an assembled version of everything needed to burn firmware
 firmware image | Built from a firmware bundle and contains the partition table, partitions, bootloader, etc.
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -17,7 +17,7 @@ $ mix local.rebar
 
 Elixir runs on top of the Erlang VM on your host as well as on your target board. What makes Nerves special is that we replace typical linux initialization mechanisms like systemd or udev with Erlang. This puts your application in a unique and powerful position of having control over the subsystem initialization as well as your application runtime.
 
-If you followed the instructions from earlier to install elixir using the instructions found on the Elixir [Installation Page](http://elixir-lang.org/install.html), you should already have Erlang installed.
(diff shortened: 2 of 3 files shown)


docs/Installation.md
```diff
@@ -41,9 +41,14 @@ Once homebrew is installed, you can install these missing utilities by running t
...
 
+For Debian-based Linux systems, you can install these utilities by running:
+```
+sudo apt-get install coreutils squashfs-tools
+```
+
 ## Nerves Bootstrap
 
-With Elixir, Erlang, and your host utilities installed, you can now add the `nerves_bootstrap` archive to your mix environment. This archive allows Nerves to bootstrap the Mix environment, ensuring that your code is properly compiled using the right cross-compiler for the target. The `nerves_bootstrap` archive also includes a new project generator, which you can use to create new Nerves projects. To install the `nerves_bootstrap` archive:
+With Elixir, Erlang, and your host utilities installed, you can now add the `nerves_bootstrap` archive to your Mix environment. This archive allows Nerves to bootstrap the Mix environment (i.e. fetch the System and Toolchain for your target), ensuring that your code is properly compiled using the right cross-compiler for the target. The `nerves_bootstrap` archive also includes a new project generator, which you can use to create new Nerves projects. To install the `nerves_bootstrap` archive:
```
mobileoverlord: > (i.e. fetch the System and Toolchain for your target)
Fetch actually happens as a side effect of compiling a system or toolchain. Bootstrap ensures that the System and Toolchain are compiled first before any other code,

ChuckLutz:
Ah, OK - I think I added that to try to align it with what was said in the Getting Started page: 
" As a first step, you need to fetch both the System and Toolchain for your Target. This task is done for you by Mix using the nerves_bootstrap utility in a special stage called precompile. This means that the first time you ask any dependencies or your application to compile, Nerves will fetch the System and Toolchain from one of our cache mirrors." 
I can take that bit out; I don't yet know all the details of the steps, besides, maybe this is too "under the hood" for someone just looking to get it installed. What do you think?

docs/Getting Started.md
```diff
@@ -22,6 +22,8 @@ host | The computer on which you are editing source code, compiling, and assembl
 target | The platform for which your firmware is built (for example, Raspberry Pi, Raspberry Pi 2, or Beaglebone Black)
 toolchain | The tools required to build code for the target, such as compilers, linkers, binutils, and C runtime
 system | A lean buildroot-based linux distribution that has been customized and cross-compiled for a particular target
+application | The Elixir (and any native, e.g. C) source code for your project. Note that this may entail more than one "application" in Elixir parlance.
```
GregMefford: I agree that it's probably a good idea to define "application" in the context of our docs, but I'm thinking maybe we should strictly use this word to refer to an Erlang/Elixir application, to reduce confusion with what is called an application in other languages. This could be a good place to point out the difference, though.

docs/Getting Started.md
```diff
@@ -22,6 +22,8 @@ host | The computer on which you are editing source code, compiling, and assembl
...
+configuration | The settings and other target-specific files that steer the firmware building process for your particular target
```
GregMefford: Should this also include the Elixir application configuration that isn't specific to Nerves?

mobileoverlord:
Hey Chuck. I think that these have been manually merged. Sorry bout that. Please take another pass and if you find edits let me know and I will merge them :bowtie:  thanks!

PR #14, opened 2016-06-15
Add app and module to switches 

TattdCodeMonkey:
# What is this?

The help docs show that `--app` and `--module` are supported, but if you try to use them you get the following error: 

> ** (Mix) Invalid option: --module

# What did I change?

Looking at the [mix.new Task](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L46) as a reference I was able to just add `app: :string, module: :string` to the `@switches` list to resolve this issue.

I tested this locally and it worked great! Cheers!


mobileoverlord:
Ah great catch. Thanks @TattdCodeMonkey

PR #15, opened 2016-06-16
Use aliases and explicit parens in mix.exs

josevalim:


fhunleth:
Thanks! This looks much nicer. I'm propagating similar changes across other Nerves packages to make them consistent.

josevalim:
Thank you! Btw, the reason I have added explicit parens to `deps` and friends is because Elixir v1.4 will warn if a variable is used as a function call. :) /cc @mobileoverlord

PR #16, opened 2016-06-21
Added central blurb about how to use a nerves system

mobileoverlord:


docs/Systems.md
```diff
@@ -1,5 +1,42 @@
...
+def deps(:dev = env) do
+  [{:my_custom_bbb_dev, ">=0.0.0"}]
+end
+
+def deps(:prod = env) do
+  [{:my_custom_bbb_prod, ">=0.0.0"}]
+end
+```
+
+Since its all done in Elixir, you can choose which configuration works best for you. Just be sure that there is only ever one `system` and one `toolchain` present when compiling your Nerves application.
```
fhunleth: Do you need to worry about only having one `toolchain` present if you only have one `system`? The rule is that only systems list a toolchain in their deps, right?

GregMefford:
Yeah, the System should be the only thing that depends on a Toolchain, so people shouldn't need to add that as a dependency themselves.

mobileoverlord:
I will update to reflect that they only need to worry about system.

mobileoverlord:
Updated

PR #19, opened 2016-07-03
Simplify installation instructions

fhunleth:
After helping someone out with installation, I think that we need a tl;dr version of the instructions. I made a pass at it in this PR. The additional text in the original seems like it would be useful somewhere, but it led to some interesting confusion. For example, `gstat` is only required on MacOS. If you don't know what it is and google it, you'll find references to Firebird database. I suspect that most people will be happy with blindly installing `coreutils`.


mobileoverlord:
Looks good

PR #22, opened 2016-07-12
DOC: Added a tip about avoiding the prompt when burning an SD card

Pacodastre:
Tip from @fhunleth in the nerves channel on slack.


mobileoverlord:
Thanks @Pacodastre :bowtie:

PR #23, opened 2016-07-12
rewrite firmware.burn to interact with fwup

mobileoverlord:


mobileoverlord:
Handles prompts better than not.

PR #26, opened 2016-07-19
supply custom vm.args if app/rel/vm.args file is not present

mobileoverlord:


mobileoverlord:
@bitwalker I have tested and verified that this will work, but I wanted to have your opinion on the strategy presented here.

The main question is: can the location of the vm.args file be variable. If so I would need a mechanism for resolving its location or this wouldn't work.

I noticed that at this point, exrm will add one to `rel/.files`, doing the check in the middle position of the relx config ensures that we are looking if the one present is the one supplied by the user or not.

With this mechanism, we could maintain our own vm.args template for Nerves applications and not worry about breaking any other projects with our focused defaults.

bitwalker:
In `exrm`, the only way to provide a custom vm.args is to either put one in `rel/vm.args`, or to use an overlay to copy/template one in to the release. Your code checks for the former, but not the latter, which you could do by checking all of the overlay `to` paths for files with a basename of `vm.args`.

In `distillery`, you will just need to check the `vm_args` setting, if non-nil, the user supplied a path of their own.

mobileoverlord:
I am going to close this for now and revisit when we switch to distillery

PR #27, opened 2016-07-26
Nerves Package Implementation

mobileoverlord:
This is a work in progress.

Included in this PR
* Consolidate toolchain and system compilers to `:nerves_package` compiler
* Remove dep on `:nerves_toolchain` and `:nerves_system`
* Add Docker provider for building package artifacts

Todo
- [x] Add tests for providers and package artifacts.
- [x] Implement local provider
- [ ] Allow individual packages to specify which provider to use
- [x] Implement provider fallbacks. For example: If HTTP fails x times fall back to docker or local


mobileoverlord:
A couple other notes, we need to perform checksums to ship with the artifact for comparison to determine if the artifact is stale or not as well as security.
I would also like to allow docker containers to persist instead of removing them right away. This will work in conjunction with the ability to define new packages like systems. We can then expose commands like `make menuconfig` and `make linux-menuconfig`

mobileoverlord:
Moved to another branch

PR #30, opened 2016-08-07
Add an Erlang distribution how-to

bendiken:
As requested by @GregMefford in [#nerves](http://slack.elixirhq.com/nerves/), here are some initial quick and dirty instructions for configuring dynamic Erlang distribution in Nerves apps:

https://github.com/bendiken/nerves/blob/erlang-distribution/docs/Erlang%20Distribution.md

See also the related `erlinit` EPMD discussion at https://github.com/nerves-project/erlinit/issues/8

### DIFF
--- a/docs/Erlang Distribution.md
+++ b/docs/Erlang Distribution.md
@@ -1,0 +1,92 @@
+# Erlang Distribution
+
+## Static Configuration
+
+The simplest way to configure Erlang distribution is to statically (i.e., at
+compile time) set the node name and cookie. This is done by editing the
+`rel/vm.args` file, which contains the command-line arguments used to launch
+the Erlang/OTP virtual machine.
+
+You would simply add `-name` and `-setcookie` parameters, as follows:
+
+```diff
+ ## Start the Elixir shell
++-name iex@nerves.local
++-setcookie nerves
+ -noshell
+ -user Elixir.IEx.CLI
+ -extra --no-halt
+```
+
+The values you give these parameters are up to you. Note that at minimum you
+will need to replace `nerves.local` in the example above with the actual
+hostname or IPv4 address that your Nerves device is configured to use or
+gets from your local DHCP server.
+
+This method of configuring distribution is therefore best suited to
+development use with a static network configuration; if you know neither the
+hostname nor IPv4 address your device will get from the local DHCP server,
+continue reading below for an alternative way to configure Erlang
+distribution.
+
+## Dynamic Configuration
+
+It is also possible to start Erlang distribution dynamically (i.e., at
+runtime) in your Nerves application, though it is a little more involved.
+
+The benefit of dynamic configuration is that there is no more need to
+hardcode the node name and cookie in `rel/vm.args` in your firmware images;
+rather, you can start distribution and bring up the Erlang node after DHCP
+autoconfiguration and app initialization is complete, using `Node.start` and
+`Node.set_cookie`. This means that you can easily run the same firmware
+image on multiple devices on the same network and have them each configure
+distribution as appropriate.
+
+Here's an outline of how:
+
+### 1. Configure a rootfs overlay
+
+In your app config, add a rootfs overlay per instructions provided in [Advanced
+Configuration](advanced-configuration.html).
+
+### 2. Override the `erlinit.config` file
+
+Copy the default `/etc/erlinit.config` file for your target into your rootfs
+overlay and customize it as per the following steps.
+
+### 3. Configure the $HOME environment variable
+
+In your `erlinit.config`, set the `$HOME` environment variable to `/tmp` (a
+writable tmpfs partition) instead of the default `/root` (the FAT32 app data
+partition on SD card).
+
+```diff
+--e LANG=en_US.UTF-8;LANGUAGE=en;ERL_INETRC=/etc/erl_inetrc
++-e LANG=en_US.UTF-8;LANGUAGE=en;ERL_INETRC=/etc/erl_inetrc;HOME=/tmp
+```
+
+This step is necessary at present in order that Erlang doesn't quit on you
+with `Cookie file /root/.erlang.cookie must be accessible by owner only` in
+attempting to autogenerate a cookie file when you call `Node.start`.
+
+The error arises because FAT32 doesn't support file modes and ownership
+metadata. We plan to address this problem in Nerves going forward by
+converting the app data partition to use another filesystem than FAT32.
+
+### 4. Start EPMD when booting
+
+In your `erlinit.config`, add a line `--alternate-exec /bin/erlwrap.sh`.
+Create the `/bin/erlwrap.sh` script in your rootfs overlay, with contents:
+
+```bash
+#!/bin/sh
+echo "Starting EPMD..."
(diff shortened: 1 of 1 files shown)


fhunleth:
Thanks for writing this up. This is sorely needed!

docs/Erlang Distribution.md
```diff
@@ -0,0 +1,92 @@
+# Erlang Distribution
+
+## Static Configuration
+
+The simplest way to configure Erlang distribution is to statically (i.e., at
+compile time) set the node name and cookie. This is done by editing the
+`rel/vm.args` file, which contains the command-line arguments used to launch
+the Erlang/OTP virtual machine.
+
+You would simply add `-name` and `-setcookie` parameters, as follows:
```
fhunleth: My preference here would be to use `-sname` or at least to document both `-sname` and `-name`. `-sname` has the advantage that the node's name can be autogenerated based on the serial number which makes things really easy for how I've been using Erlang distribution. The main problem with `-sname` is that it works best if the DHCP server can register the name with DNS, and it's hit or miss whether people's routers support this. @mobileoverlord: Which do you use now? `-name` or `-sname`? And is it only for debug or have you used Erlang distribution in production?

Regarding use of `nerves.local`, I had thought that that would only work if you had an mDNS service running on your Nerves device. Or is this not true?

docs/Erlang Distribution.md
```diff
@@ -0,0 +1,92 @@
...
+
+### 3. Configure the $HOME environment variable
+
+In your `erlinit.config`, set the `$HOME` environment variable to `/tmp` (a
+writable tmpfs partition) instead of the default `/root` (the FAT32 app data
+partition on SD card).
+
+```diff
+--e LANG=en_US.UTF-8;LANGUAGE=en;ERL_INETRC=/etc/erl_inetrc
++-e LANG=en_US.UTF-8;LANGUAGE=en;ERL_INETRC=/etc/erl_inetrc;HOME=/tmp
```
fhunleth: I'd like to fix the `/root` filesystem to be EXT4 before we merge this. I don't like pointing `HOME` to `/tmp` in the official docs even though I completely understand why you did this. @mobileoverlord - we should talk about timing this update unless you don't want it. The hard part will be that `/root` can't be formatted by `fwup`. We'll need to format it on the first boot and whenever it gets corrupt like the `nerves_ev3_example` does.

docs/Erlang Distribution.md
bendiken: I had understood that `-sname` is only for nodes on the same host. Had I misunderstood that?

As for `.local`, there isn't anything special about the extension as such. You can name your local hosts with that extension with or without mDNS. We need some example name to use in the docs, so I used that.

docs/Erlang Distribution.md
bendiken: OK... without this bit, these instructions can't work, at present. Could we do something else such as layer a tmpfs over /root? I wouldn't like to leave this pull open for (potentially) months.

docs/Erlang Distribution.md
fhunleth: Out of curiousity, how does that work, then? If I have a Nerves node and give it the name `nerves.local`, then how does my laptop do the DNS resolution when I refer to `nerves.local` from it? I thought that the only two options were DNS and mDNS.

`-sname` can be used between hosts. I actually use it more frequently since I don't want to deal with the fully qualified names. The hostname part of the node name has always felt like a dynamic so I never wanted to lock it down to a fully qualified name. It occurs to me that your PR to change the node name at runtime might just be another way of addressing the same issue.

docs/Erlang Distribution.md
```diff
@@ -0,0 +1,92 @@
...
+with `Cookie file /root/.erlang.cookie must be accessible by owner only` in
+attempting to autogenerate a cookie file when you call `Node.start`.
+
+The error arises because FAT32 doesn't support file modes and ownership
+metadata. We plan to address this problem in Nerves going forward by
+converting the app data partition to use another filesystem than FAT32.
+
+### 4. Start EPMD when booting
+
+In your `erlinit.config`, add a line `--alternate-exec /bin/erlwrap.sh`.
```
fhunleth: I think this section should just say add the line `System.cmd("/usr/lib/erlang/bin/epmd", ["-daemon"])` to your OTP Application's `start/2` function. That is until we have a nicer way to supervise or launch it. It's simpler and avoids the shell scripting and erlinit.config steps. Fwiw, I got excited when I saw that `erlexec` has a `-start_epmd true` option, but it turns out that it only really starts `epmd` if `-sname` or `-name` are specified as well.

docs/Erlang Distribution.md
fhunleth: There's an expectation that `/root` persists. 

Another option would be to split this PR into what's supported today and what's supported once the `/root/erlang.cookie` issue is fixed. I don't think that it will be months, though.

docs/Erlang Distribution.md
bendiken: I add `nerves.local` (and similar names) in my laptop's `/etc/hosts`, which enables me to connect to the Nerves nodes using `iex`.

docs/Erlang Distribution.md
bendiken: Will do, once I have a chance to test that that actually works.

GregMefford:
Wow, sorry this has been sitting for so long! 😞 

I was planning to work on merging the `hello_wifi` and `hello_network` examples together and also adding distribution support so that we'll have a working example posted. I'll be sure to incorporate this PR as a part of that effort.

My plan was to use the `NetworkManager`-style code as Justin used on his [Kegerator](https://github.com/mobileoverlord/kegerator/blob/master/apps/kegerator/lib/kegerator/network_manager.ex) project, which gets around the issue with `/root/erlang.cookie` by [setting an initial cookie in `vm.args`](https://github.com/mobileoverlord/kegerator/blob/master/apps/kegerator/rel/vm.args#L1).

@fhunleth and @mobileoverlord: Are you still thinking you want to make the `/root` partition `EXT4`? I think we can land this PR using the method above, but I just wanted to make sure we didn't lose track of that if it's something we wanted to do for other reasons.

fhunleth:
I personally think that EXT4 is a requirement for `/root` in any project. The hangup has been getting the reformat-on-corruption (otherwise known as format-on-first-use) integrated in a good way. Now that `nerves_runtime` exists, there's a route. EXT4 support should be its own issue if it's not already somewhere else.

mobileoverlord:
@bendiken Sorry that this has drug on so long. A lot has changed.

Restarting the net kernel can be be handled these days by updating a couple things. You can use [Bootloader]() in your project, which will allow you to list [nerves_init_net_kernel]() to the init applications list for bootloader. This will handle monitoring the network interface through [system_registry]() for when changes happen on the [nerves_network]() interface.

We are in the process of updating our documentation to reflect a lot of these new structure changes, but, if you get the latest nerves_bootstrap archive, you can use it to generate a new project and see how the new layout works.

I really appreciate all the work you have done with nerves so far. Thanks for your involvement, I will be more attentive in the future to make sure that we don't wait too long on PR's before getting them addressed. Thanks 😄

PR #31, opened 2016-08-08
ui config in fw project and also OTP-configuration

ulve:
The documentation for how to crete a UI is somewhat out of date. At the momen sqlite does not work and is not needed also brunch is not needed. Configuration for the ui app in the fw app is missing and also configuration for the OTP application is missing. fw.ex must also have a child worker that sets up networking.

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -1,4 +1,4 @@
-# User Interfaces
+  # User Interfaces
 
 ## Phoenix Web Interfaces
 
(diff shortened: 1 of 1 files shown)


mobileoverlord:
It would be really nice to see the ecto sqlite adapter be updated for ecto 2.0, but you are right, this needs to be added. Thanks!

:bowtie:

GregMefford:
Thank you for taking the time to improve this, @ulve! :balloon:

PR #35, opened 2016-08-14
Start docs for firmware and firmware.burn

fhunleth:
@mobileoverlord - hopefully, you can fix up anything that I got wrong. I just wanted to get something started once I noticed that we were lacking `mix` docs for these.


mobileoverlord:
❤️ :bowtie:

PR #43, opened 2016-09-07
Add support for Windows

michaelkschmidt:
For reasons I don't understand, System.cmd("which",[]) just hangs on Windows

### DIFF
--- a/lib/mix/tasks/firmware.burn.ex
+++ b/lib/mix/tasks/firmware.burn.ex
@@ -75,6 +75,8 @@ defmodule Mix.Tasks.Firmware.Burn do
            ask_pass = System.get_env("SUDO_ASKPASS") || "/usr/bin/ssh-askpass"
            System.put_env("SUDO_ASKPASS", ask_pass)
            {"sudo", ["fwup"] ++ args}
+        {_, :nt} ->
+           {"fwup", args}
         {_, type} ->
           raise "Unable to burn firmware on your host #{inspect type}"
       end
--- a/lib/mix/nerves/utils.ex
+++ b/lib/mix/nerves/utils.ex
@@ -12,7 +12,14 @@ defmodule Mix.Nerves.Utils do
   end
 
   def check_requirements do
-    case System.cmd("which", ["mksquashfs"]) do
+    {_, type} = :os.type
+
+    which_or_where = case type do
+                        :nt -> "where"
+                        _ -> "which"
+                     end
+                     
+    case System.cmd(which_or_where, ["mksquashfs"]) do
       {_, 0} -> nil
       _ -> Mix.raise """
       Squash FS Tools are required to be installed on your system.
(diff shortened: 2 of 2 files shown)


lib/mix/tasks/firmware.burn.ex
```diff
@@ -75,6 +75,8 @@ defmodule Mix.Tasks.Firmware.Burn do
            ask_pass = System.get_env("SUDO_ASKPASS") || "/usr/bin/ssh-askpass"
            System.put_env("SUDO_ASKPASS", ask_pass)
            {"sudo", ["fwup"] ++ args}
+        {_, :nt} ->
+           {"fwup", args}
```
fhunleth-lkc: Is `fwup` working fine on Windows? I assume that you're using 0.9.0-dev. I'm planning on making an official `fwup` release really soon, but if there are any annoyances, now would be a good time for me to fix them.

michaelkschmidt: It has been working great!  I cross compiled it from a Linux VM, and haven't had a single issue.  It even detects the SD Card correctly :+1:

fhunleth-lkc: Super! @GregMefford wrote the SD Card detection code and will be thrilled to hear.

fhunleth:
Since I don't think that any of us said this yet, thank you so much for the Windows patches. We're in the process of catching up from ElixirConf, but we'll get to this patch soon!

Also, I just officially released fwup v0.9.0 which has Windows support so we're getting closer.

michaelkschmidt:
No problem.  When you get back to the patches, let me know if there is something you would prefer changed.  I tried hard to get tar to extract the .tar.xz files, but I kept running into weird interactions with the MinGW/Cygwin path.  7zip is a bit of a punt, but at least its easily installed via Chocolatey.

Speaking of Chocolatey, do you plan to distribute fwup on it?  I would be glad to try to help.  The squashfs tools should probably be put on there as well, too, but I'm not sure who would be the appropriate person to do so.

fhunleth:
Adding fwup and the squashfs tools to Chocolatey would be ideal. I've never submitted anything to Chocolatey before, but now that you mention it, I'd like to do it. I'm hanging it up for the night, but let's chat over Slack on this and see if we can get these included.

lib/mix/tasks/firmware.burn.ex
GregMefford: Very exciting when things just work!

mobileoverlord:
This is great, thanks @michaelkschmidt I am going to merge this in and add it to the new pkg branch coming in.

Thanks! :bowtie:

fhunleth:
@michaelkschmidt Maybe I haven't seen it yet, but how did you handle toolchains on Windows? Did you build your own?

michaelkschmidt:
The short answer is I built my own toolchain.

I was hoping to get your feedback on the long answer--I'll ping you on slack

PR #46, opened 2016-09-18
Add support for Elixir >= 1.4.0-dev

wsmoak:
Similar to a01596ec59, allow use of Elixir 1.4.0-dev (the version currently on master.)

This prevents a warning every time mix is used while nerves.archive is installed.


mobileoverlord:
@wsmoak I added support for this in this commit manually.
113c06369d2c492a39dfb260078fc3488487b044

Thanks!
:bowtie:

mobileoverlord:
this version of nerves_bootstrap may not function properly when used with existing projects.

PR #50, opened 2016-09-25
Use NERVES_TARGET when building examples

pkazmier:
Small doc changes:

1. Added a note about specifying the target via environment variable to example at the end of the page. This tripped me as a new user following along as I ended up building to the wrong target.

2. Reduced the heading level of questions in the FAQ document to maintain better visual balance on the page while planning for future sections that can be used to group the questions as the document grows.

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -127,11 +127,11 @@ $ mix firmware.burn -y
 
 ## Nerves Examples
 
-To get up and running quickly, you can check out our collection of example projects:
+To get up and running quickly, you can check out our collection of example projects. Be sure to set your `NERVES_TARGET` environment variable appropriately as the examples use this to determine the target when building firmware. Visit the [Targets Page](targets.html) for more information on what target name to use for the boards that Nerves supports.
 ```
 $ git clone https://github.com/nerves-project/nerves-examples
 $ cd nerves-examples/blinky
-$ mix deps.get && mix firmware
+$ NERVES_TARGET=rpi3 sh -c "mix deps.get && mix firmware"
 ```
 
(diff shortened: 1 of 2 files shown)


docs/Getting Started.md
```diff
@@ -127,11 +127,11 @@ $ mix firmware.burn -y
 
 ## Nerves Examples
 
-To get up and running quickly, you can check out our collection of example projects:
+To get up and running quickly, you can check out our collection of example projects. Be sure to set your `NERVES_TARGET` environment variable appropriately as the examples use this to determine the target when building firmware. Visit the [Targets Page](targets.html) for more information on what target name to use for the boards that Nerves supports.
 ```
 $ git clone https://github.com/nerves-project/nerves-examples
 $ cd nerves-examples/blinky
-$ mix deps.get && mix firmware
+$ NERVES_TARGET=rpi3 sh -c "mix deps.get && mix firmware"
```
GregMefford: I'm concerned that this little "shell magic" might be intimidating to newcomers. As an alternative, how about `NERVES_TARGET=rpi3 mix do deps.get, firmware`. This has the additional benefit of introducing the existence of the `mix do` command, which people may not have encountered yet.

GregMefford:
Sorry for the big red X in your face! I was playing with the new GitHub Code Review feature and didn't realize the "request changes" option would look so harsh.

I think this change is mostly great and I don't mean to give you that gut reaction to your contribution being rejected. 🎈

docs/Getting Started.md
pkazmier: Sure. I have not encountered the `mix do` either. I like that much better. Will resubmit tomorrow.

PR #51, opened 2016-10-08
Use NERVES_TARGET when building examples

pkazmier:
Added a note about specifying the target via environment variable to example at the end of the page.


GregMefford:
Thanks, @pkazmier! For the record, it looks like this is an update to #50.
Did you also want to decrease the heading levels like you had in that PR?

PR #52, opened 2016-10-08
Reduced header level of FAQ questions

pkazmier:
The questions are currently a level 1 heading, which is too large given the length of the questions. By changing these to level 3 headings, the visual balance of the page is improved.

By using level 3 headings for the questions, level 2 headings are available for use later to group questions once the FAQ becomes large enough. In addition, these future level 2 section headers will render in the navigation tree.


GregMefford:
I agree, thanks! 🚀 ✨

PR #53, opened 2016-10-15
Update to latest ex_doc release

wsmoak:
ExDoc has had a few updates recently. 

This updates the dependencies so that Nerves docs will be built with the latest ExDoc and Earmark.

(I believe support for epub is included, but I haven't tried it out yet.)


mobileoverlord:
Thanks Wendy!

Issue #61, opened 2016-12-16
Add update mechanism to nerves_bootstrap


mobileoverlord:
Nerves Bootstrap needs to be able to inform the user that there is an update available for their version. we should create a mix task which installs the latest bootstrap version.

ghitchens:
this closes the mix task question but not the inform, did we decide not to inform, or does it do that now?

mobileoverlord:
Well, I guess that this inevitably solves the task at hand. We would need to implement an API on our site to provide bootstrap update requests.

ghitchens:
I guess the notification part is handled by things like ~> "0.2" in mix.exs of the app using it.  I'm ok closing it.

PR #64, opened 2017-01-10
Add `mix local.nerves` to update local nerves archive

tverlaan:
No checking if Nerves is outdated, but at least an easy way to update the local archive. I 'borrowed' the code from Phoenix.

Solves #61


mobileoverlord:
Thanks! 🎊

PR #66, opened 2017-01-30
make output of images configurable

ConnorRigby:
in your mix config you can have something like:
```elixir
  def project do
    [app: :my_app,
     version: "1.2.3",
     target: "rpi3",
     archives: [nerves_bootstrap: "~> 0.2.0"],
     build_embedded: Mix.env == :prod,
     start_permanent: Mix.env == :prod,
     build_path:  "../../_build/#{target(Mix.env)}",
     deps_path:   "../../deps/#{target(Mix.env)}",
     images_path: "../../images", # THIS
     config_path: "../../config/config.exs",
     lockfile: "../../mix.lock",
     aliases:     [],
     deps:        deps()
   ]
  end
```

### DIFF
--- a/lib/mix/tasks/firmware.burn.ex
+++ b/lib/mix/tasks/firmware.burn.ex
@@ -42,7 +42,9 @@ defmodule Mix.Tasks.Firmware.Burn do
     config = Mix.Project.config
     otp_app = config[:app]
     target = config[:target]
-    images_path = config[:images_path] || config[:build_path] || "_build/images"
+    images_path =
+      (config[:images_path] || Path.join([Mix.Project.build_path, "nerves", "images"]))
+      |> Path.expand
 
     System.get_env("NERVES_SYSTEM") || raise """
       Environment variable $NERVES_SYSTEM is not set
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/firmware.burn.ex
```diff
@@ -42,6 +42,7 @@ defmodule Mix.Tasks.Firmware.Burn do
     config = Mix.Project.config
     otp_app = config[:app]
     target = config[:target]
+    images_path = config[:images_path] || config[:build_path] || "_build/images"
```
mobileoverlord: Lets make this 
```
images_path = config[:images_path] || Path.join([Mix.Project.build_path, "nerves", "images"])
```
```

Comment (mobileoverlord) [inline lib/mix/tasks/firmware.burn.ex]: actually we would need to expand it if it is passed.

images_path = 
  (config[:images_path] || Path.join([Mix.Project.build_path, "nerves", "images"]))
  |> Path.expand

File: lib/mix/tasks/firmware.burn.ex [hunk 2]
```diff
@@ -51,7 +52,7 @@ defmodule Mix.Tasks.Firmware.Burn do
       Environment variable $NERVES_TOOLCHAIN is not set
     """
 
-    fw = Path.join(File.cwd!, "_images/#{target}/#{otp_app}.fw")
+    fw = Path.join(File.cwd!, "#{images_path}/#{target}/#{otp_app}.fw")
```
mobileoverlord: I think that fwup will create the dir if it does not exist but we should probably `File.mkdir_p` just incase.

mobileoverlord: also, in this case we would not need target anymore since the default build path would contain it.

ConnorRigby: ah good point.

lib/mix/tasks/firmware.burn.ex
ConnorRigby: ah didn't see `Mix.Project.build_path` nice

mobileoverlord:
Thanks :bowtie:

PR #72, opened 2017-02-10
Return error when tar command isn't successful

texjim:
Whenever we are pulling down tar files, if the connection drops the tar file may be incomplete.  To make this a bit more robust, I added code to catch the return code from the tar command.   If it's not 0, then return an error.  Otherwise keep on trucking.

Also, I created a unit test to test this, it passes in a .fw file instead of a tar ball which obviously should throw an error.   The .fw file is fairly large, so I have it skipping it.   Any feedback is welcome, as I wasn't sure what the best strategy for a test would be.


mobileoverlord:
👍 this is great! Thanks!

PR #77, opened 2017-02-23
Add experimental host shell accessible through Erlang's job control mode from IEx

electricshaman:
This is an experimental host shell emulator that is available through the job control mode in Erlang's shell.  Right now it's basically a barebones ripoff of IEx that executes commands against a port that spawns `/bin/sh` on the host.  There are a few obvious loose ends that still need tying up, but this should be enough to get us going so we can continue to collectively iterate on it over time.  The workflow looks something like this at the moment:

```
$ iex -S mix
Erlang/OTP 19 [erts-8.2] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Compiling 1 file (.ex)
Interactive Elixir (1.4.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> (press Ctrl+G)
User switch command
 --> s 'Elixir.Nerves.Host'
 --> c
Interactive Host Shell
host[1]> ls -al

total 32
drwxr-xr-x  11 jeff  staff  374 Feb 13 19:46 .
drwxr-xr-x   4 jeff  staff  136 Feb  8 16:03 ..
-rw-r--r--   1 jeff  staff  503 Feb  8 16:03 .gitignore
-rw-r--r--   1 jeff  staff  474 Feb  8 16:03 README.md
drwxr-xr-x   3 jeff  staff  102 Feb  8 16:39 _build
drwxr-xr-x   3 jeff  staff  102 Feb  8 16:03 config
drwxr-xr-x   3 jeff  staff  102 Feb  9 00:07 deps
drwxr-xr-x   4 jeff  staff  136 Feb 14 15:19 lib
-rw-r--r--   1 jeff  staff  789 Feb  9 00:06 mix.exs
-rw-r--r--   1 jeff  staff  123 Feb  9 00:07 mix.lock
drwxr-xr-x   4 jeff  staff  136 Feb  8 16:03 test

host[2]> (press Ctrl+G)
User switch command
 --> j
   1  {erlang,apply,[#Fun<Elixir.IEx.CLI.1.112225073>,[]]}
   2* {'Elixir.Nerves.Host',start,[]}
 --> c 1
Interactive Elixir (1.4.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)>
```

Feedback definitely welcome.  This will be an ongoing experiment.


fhunleth:
This is going to be so useful! 

I wish that there were some way to make Ctrl+C work sensibly or allow raw mode programs to work (ones that use ncurses, etc.), but I think that's ok. I wonder if some help text when starting up the shell would be useful to remind people that there are limitations.

Also, you probably noticed that I've disabled quite a few shell commands on Nerves. We should add some back to make this useful. If you don't see how to do it with Busybox in `nerves_system_br`, let me know.

Regarding naming, would you be ok with calling it `Nerves.Shell` or something similar? The word `Host` has gotten overloaded.

electricshaman:
ncurses and Ctrl+C are at the top of my list too.  That would be amazing.  I'm still hopeful we can get it there.  The other thing I'm not giving up on yet is the ability to send single characters.  Right now there is a hard requirement on newline termination.  That's going to require digging deeper into how Erlang's `tty_sl` port driver works.  

I'm good with `Nerves.Shell`.  I do think that's more intuitive in this context.  Earlier we had been trying to use it in conjunction with the new host target features so that's how we ended up with `Host`.  @mobileoverlord You good with this rename too?

mobileoverlord:
I am so excited to start using this ❤️

fhunleth:
Me too!!! Thanks!

GregMefford:
For people finding this later, I feel like we should mention that this has been moved to `nerves_runtime` now.
See #81 and https://github.com/nerves-project/nerves_runtime/pull/1

PR #81, opened 2017-02-28
Move experimental shell emulator to runtime

electricshaman:
This is the first of 2 PRs to move the experimental shell emulator to the new runtime library.


mobileoverlord:
see you in runtime! How we hardly knew ye 😄

PR #82, opened 2017-02-28
Include nerves_runtime

mobileoverlord:
sets runtime false on compile time deps and includes nerves_runtime in targets deps

### DIFF
--- a/bootstrap/lib/mix/tasks/nerves/new.ex
+++ b/bootstrap/lib/mix/tasks/nerves/new.ex
@@ -3,7 +3,11 @@ defmodule Mix.Tasks.Nerves.New do
   import Mix.Generator
 
   @nerves Path.expand("../../../..", __DIR__)
-  @version "0.2"
+
+  @nerves_vsn "0.4.0"
+  @bootstrap_vsn "0.2"
+  @runtime_vsn "0.1"
+
   @requirement Mix.Project.config[:elixir]
   @shortdoc "Creates a new Nerves application"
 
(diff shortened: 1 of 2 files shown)


bootstrap/lib/mix/tasks/nerves/new.ex
```diff
@@ -3,7 +3,11 @@ defmodule Mix.Tasks.Nerves.New do
   import Mix.Generator
 
   @nerves Path.expand("../../../..", __DIR__)
-  @version "0.2"
+
+  @nerves_vsn "0.4.0"
```
GregMefford: Is there a reason we're constraining the `nerves` version to the minor release, but `bootstrap` and `runtime` only to the major version? I'm thinking most people would likely only want bugfixes/patches applied automatically to their projects, as opposed to minor/feature releases.

bootstrap/templates/new/mix.exs
```diff
@@ -59,7 +59,8 @@ defmodule <%= app_module %>.Mixfile do
   # Specify target specific dependencies
   def deps("host"), do: []
   def deps(target) do
-    [{:"nerves_system_#{target}", ">= 0.0.0"}]
+    [{:nerves_runtime, "~> <%= runtime_vsn %>"},
+     {:"nerves_system_#{target}", ">= 0.0.0", runtime: false}]
```
GregMefford: Again, I wonder if there's a down-side to specifying a version more specifically here, so that the project owner clearly sees that they can/should choose what version to use for their project.

mobileoverlord: The system repositories are synced on version now, but they could diverge. specifying a version in this case would blanket it across all system repos.

bootstrap/lib/mix/tasks/nerves/new.ex
mobileoverlord: Our update mechanisms for bootstrap and nerves are more pronounced now, we could lock the others down

GregMefford: I think it's worth specifying them as all `x.y.z` unless there's a reason not to, so that we cause the least amount of surprise and let people choose when they want to upgrade to the next feature release.

bootstrap/templates/new/mix.exs
GregMefford: Good point, but we still do release them in lock-step, with only the patch-level being potentially different, right? For example, we'd set the constraint to `"~> 0.10.0"` today.

mobileoverlord: Yeah, today they are in lock step, but its not required that they are.

mobileoverlord: Alright, you got me. I've got no more excuses.

mobileoverlord:
@GregMefford I locked the deps down and changed the system dep to start @ ~> `0.10.0`

Issue #87, opened 2017-03-05
Documentation: Getting Started section mix firmware generates error message about missing release config file


inactivist:
### Environment
* Erlang version: 19
* Elixir version (elixir -v): 1.3.4
* Operating system: Ubuntu Linux 15.10 x64 with Kernel 4.4.8-040408-generic
* Nerves Environment Info:
```
$ mix nerves.env --info
==> distillery
Compiling 18 files (.ex)
Generated distillery app
==> nerves
Compiling 25 files (.ex)
Generated nerves app
==> nerves_system_br
Generated nerves_system_br app
==> nerves_toolchain_ctng
Compiling 2 files (.ex)
Generated nerves_toolchain_ctng app
==> nerves_toolchain_arm_unknown_linux_gnueabihf
Compiling 1 file (.ex)
Generated nerves_toolchain_arm_unknown_linux_gnueabihf app
==> nerves_system_rpi2
Generated nerves_system_rpi2 app
==> hello_nerves
---------------------------
Nerves Environment Packages
---------------------------

Pkg:      nerves_system_br
Vsn:      0.9.4
Type:     system_platform
Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.Local, []}]

Pkg:      nerves_system_rpi2
Vsn:      0.10.0
Type:     system
Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.Local, []}]

Pkg:      nerves_toolchain_arm_unknown_linux_gnueabihf
Vsn:      0.9.0
Type:     toolchain
Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.HTTP, []}]

Pkg:      nerves_toolchain_ctng
Vsn:      0.8.0
Type:     toolchain_platform
Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.Local, []}]

Nerves Loadpaths Start
Nerves Env Start
Nerves Env End
----------------------------
Nerves Environment Variables
----------------------------
target:     rpi2
toolchain:  /home/inactivist/.nerves/artifacts/nerves_toolchain_arm_unknown_linux_gnueabihf-0.9.0.linux-x86_64
system:     /home/inactivist/.nerves/artifacts/nerves_system_rpi2-0.10.0.arm_unknown_linux_gnueabihf
app:        /home/inactivist/m/devel/mine/nerves-tests/hello_nerves

[...]

PR #80, opened 2017-02-27
Update docs for Nerves 0.5.0 / Bootstrap 0.3.0 release

GregMefford:
Prepping documentation for release of `nerves` 0.5.0 and `nerves_bootstrap` 0.3.0.

- [x] Getting Started page
- [x] Installation page
- [x] Systems page
- [x] Targets page
- [x] User Interfaces page
- [x] Advanced Configuration page
- [x] FAQ page
- [x] How to use the Docker provider
  - [x] in Installation
  - [x] in Systems

PR #88, opened 2017-03-05
Fix Making Firmware section

inactivist:
Fix Making Firmware section

This is to address issue #87.

- Add section on generating a release config file prior to running
   `mix firmware` command

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -100,7 +100,24 @@ You'll notice that subsequent calls to `compile` will not fetch or build the sys
 
 ## Making Firmware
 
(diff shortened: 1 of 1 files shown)


inactivist:
Please provide feedback if edits are insufficient or confusing.  I tried to keep it as simple as possible, just enough to keep newbies from seeing the error message.

GregMefford:
Thank you for taking the time to help improve the documentation, it's a huge help! 🎈 🚀 
We're just about to publish a major update to the docs, but I'll be sure to get this merged where it makes sense once that's done.

GregMefford:
This was merged manually in 20c866d along with #80. Thanks!

PR #89, opened 2017-03-06
Nerves v0.5.0 & Nerves.Bootstrap v0.3.0

mobileoverlord:
Lets do this!

Action packed release


mix.exs
```diff
@@ -6,8 +6,8 @@ defmodule Nerves.Mixfile do
      name: "Nerves",
      source_url: "https://github.com/nerves-project/nerves",
      homepage_url: "http://nerves-project.org/",
-     version: "0.5.0-dev",
-     archives: [nerves_bootstrap: "~> 0.3 or 0.3.0-dev"],
+     version: "0.5.0",
+     archives: [nerves_bootstrap: "~> 0.3.0"],
      elixir: "~> 1.2.4 or ~> 1.3.2 or ~> 1.4",
```
GregMefford: I think this should be just `"~>1.4.0"` now.

bootstrap/mix.exs
```diff
@@ -3,7 +3,7 @@ defmodule Nerves.Bootstrap.Mixfile do
 
   def project do
     [app: :nerves_bootstrap,
-     version: "0.3.0-dev",
+     version: "0.3.0",
      elixir: "~> 1.4",
```
GregMefford: We should probably make this `"~> 1.4.0"`

mix.exs
mobileoverlord: Yup you're right

bootstrap/mix.exs
mobileoverlord: okay, but if all this lock down gets weird, I am totally going to refer back to this conv 😺

PR #95, opened 2017-03-11
Doc/fix config loading and live reload disabling

frahugo:
LiveReload and `ui` config being compiled last was making the app crash on my rPi2.


mobileoverlord:
I was just discussing this with Greg last week. Sorry for the delay, recovering from EEF :bowtie: We should bring this up for discussion for the new project generator on how it should handle configuration in an umbrella project. I am starting to believe that the default should behave no differently than path dependencies and that the app should only use its own config.exs

PR #98, opened 2017-03-19
docs: declare worker in lib/fw/application.ex

salzig:
For me it looks like thats the current place where the worker should be declared. fw.ex only contains an empty module, start is handled in fw/application.ex.


GregMefford:
Good catch! I think this problem crept in as a result of structural changes in the Nerves project generator over time.

Thanks for taking the time to submit a fix when you noticed a problem with the documentation; it's very important to us! 🎈 🚀 

You may be interested is also see the `hello_phoenix` example project, though it still follows that old structure and should probably be updated to match this new convention: https://github.com/nerves-project/nerves-examples/blob/master/hello_phoenix/apps/fw/lib/fw.ex

PR #99, opened 2017-03-19
Bump doc and distillery deps in mix.lock

fhunleth:
This clears up a ton of warnings when generating the documentation.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -26,9 +26,9 @@ defmodule Nerves.Mixfile do
 
   defp deps do
     [
-      {:distillery, "~> 1.0"},
-      {:earmark, "~> 1.0", only: :dev},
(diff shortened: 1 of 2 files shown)


GregMefford:
Awesome, thanks! I remember wondering why `ex_doc` was generating all those deprecation warnings but must have checked for an updated release just before `0.15.0` was out. It looks like they forgot to actually tag it in GitHub, but there it is on Hex. 😉

GregMefford:
I figure we might as well bump the version requirements in `mix.exs` so people will be sure to get the ones we want them to have, rather than relying on `mix.lock`. Does that make sense, or is there a reason to avoid doing that?

mobileoverlord:
Ah yes, those pesky docs errors, finally solved :) Thanks!

PR #100, opened 2017-03-19
Spell out mix environment

fhunleth:
Granted that this is some bikeshedding, but the environment was referred to as the "Mix environment" in the `Installation.md` doc. I could only find the shortened version of "env" in code rather than English docs. To me, this looks better and more consistent.


GregMefford:
I agree - thanks for taking the time to be meticulous! 🚀

PR #102, opened 2017-03-23
Add note on specifc mix nerves.release.init task

cblock:
I just ran into the issue of using `mix release.init` generator instead of `mix nerves.release.init`. Both generate a config.exs. So it was hard for me to trace back the resulting errors later on to this cause. Hopefully the note helps others avoid this trap.


GregMefford:
Thanks for taking the time to improve the documentation! 🎈 🚀

PR #103, opened 2017-03-23
Update section on customizing mix.exs

cblock:
Add `:nerves_runtime` dependency to `deps` function as this is now part of the generated mix.exs


GregMefford:
Awesome! Thanks so much for taking the time to improve the docs! 🎈 🚀

PR #104, opened 2017-03-24
Updates whitespace in bootstrap files

influenza:
When using `mix nerves.new some_project` I noticed the generated `mix.exs` file had some extra newlines and trailing whitespace.

After updating the relevant template, the output now looks like this:

Additionally I removed some trailing whitespace from the `app_name.ex` template.

### DIFF
--- a/bootstrap/templates/new/lib/app_name.ex
+++ b/bootstrap/templates/new/lib/app_name.ex
@@ -1,3 +1,3 @@
 defmodule <%= app_module %> do
-  
+
 end
--- a/bootstrap/templates/new/mix.exs
+++ b/bootstrap/templates/new/mix.exs
@@ -12,14 +12,14 @@ defmodule <%= app_module %>.Mixfile do
      version: "0.1.0",
      elixir: "<%= elixir_req %>",
      target: @target,
-     archives: [nerves_bootstrap: "~> <%= bootstrap_vsn %>"],<%= if in_umbrella do %>
-     deps_path: "../../deps/#{@target}",
+     archives: [nerves_bootstrap: "~> <%= bootstrap_vsn %>"],
+<%= if in_umbrella do %>     deps_path: "../../deps/#{@target}",
      build_path: "../../_build/#{@target}",
      config_path: "../../config/config.exs",
-     lockfile: "../../mix.lock",<% else %>
-     deps_path: "deps/#{@target}",
-     build_path: "_build/#{@target}",<% end %>
-     build_embedded: Mix.env == :prod,
+     lockfile: "../../mix.lock",
+<% else %>     deps_path: "deps/#{@target}",
+     build_path: "_build/#{@target}",
+<% end %>     build_embedded: Mix.env == :prod,
      start_permanent: Mix.env == :prod,
      aliases: aliases(@target),
      deps: deps()]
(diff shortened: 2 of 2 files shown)


bootstrap/templates/new/mix.exs
```diff
@@ -12,14 +12,14 @@ defmodule <%= app_module %>.Mixfile do
      version: "0.1.0",
      elixir: "<%= elixir_req %>",
      target: @target,
-     archives: [nerves_bootstrap: "~> <%= bootstrap_vsn %>"],<%= if in_umbrella do %>
-     deps_path: "../../deps/#{@target}",
+     archives: [nerves_bootstrap: "~> <%= bootstrap_vsn %>"],
+<%= if in_umbrella do %>     deps_path: "../../deps/#{@target}",
```
influenza: While this isn't great for readability in the template itself, I think that creating clean output is more important.

bootstrap/templates/new/lib/app_name.ex
```diff
@@ -1,3 +1,3 @@
 defmodule <%= app_module %> do
-
```
electricshaman: Good catch.

GregMefford:
While we're at it, I think we should add newlines before and after the mix env information output to offset it better: https://github.com/nerves-project/nerves/pull/104/files#diff-ec9bd065f81ad68f6af89a05b8198e24R5

adkron:
:+1: :100:

GregMefford:
I was working on a final review to get this merged and I had this nagging feeling the Ruby's template language has a way to trim whitespace so I went to check whether Elixir does, too. It turns out you can [set the `trim: true` option](https://hexdocs.pm/eex/EEx.html) to do this, which would make both the template and the result look nice.

Before trying that change [here](https://github.com/nerves-project/nerves/blob/master/bootstrap/lib/mix/tasks/nerves/new.ex#L233), I wanted to validate that I see the same behavior as you describe, so I can confirm it's fixed. I wasn't able to see the same result when I generate a new project, whether it's in an umbrella or not. Can you confirm what version if Elixir you're using and the exact steps to cause this file to get generated?

Here's what I see with Vim set to `:set listchars=eol:$,tab:>-,trail:~,extends:>,precedes:<`:
nerves_template

influenza:
Version information:
```
➜  nerves git:(2017-03-23-Whitespace-Updates) elixir --version
Erlang/OTP 19 [erts-8.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.4.2
➜  nerves git:(2017-03-23-Whitespace-Updates) mix --version
Erlang/OTP 19 [erts-8.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Mix 1.4.2
```

When generating the files on my laptop (I'm traveling currently) I do not see the whitespace problem. I suspect my home machine may have an older version of Elixir.

GregMefford:
Thanks for following-up, and for taking the time to put in this PR initially! 🎈 

I went ahead and manually added the whitespace to `mix.exs` around the environment output, which was independent of the original issue but didn't want to forget about it.
486ecd565eda8f5f806c6eb21e7fb6591522d6b2

PR #105, opened 2017-03-24
Add comment explaining SUDO_ASKPASS

influenza:
This environment variable is read in
`nerves/lib/mix/tasks/firmware.burn.ex` but not documented anywhere. On
Arch Linux systems, `ssh-askpass` is not located in `/usr/bin` (or even
on the path!). This commit updates the documentation with a note
explaining how to solve this situation.


fhunleth:
This is definitely helpful. Thanks for the update.

Issue #110, opened 2017-04-14
Invalid generated docker id 


alfert:
### Environment
* Erlang version (erl -eval 'erlang:display(erlang:system_info(otp_release)), halt().'  -noshell): 19
* Elixir version (elixir -v): 1.4
* Operating system: MacOs Darwin Kernel Version 15.6.0
* Nerves Environment Info:
  (mix nerves.env --info) or
  General info about target / system / toolchain
```
|nerves_bootstrap| Environment Package List

  Pkg:      nerves_system_br
  Vsn:      0.10.0
  Type:     system_platform
  Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.Docker, []}]

  Pkg:      nerves_toolchain_armv6_rpi_linux_gnueabi
  Vsn:      0.10.0
  Type:     toolchain
  Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.HTTP, []}]

  Pkg:      nerves_toolchain_ctng
  Vsn:      0.9.0
  Type:     toolchain_platform
  Provider: [{Nerves.Package.Providers.HTTP, []}, {Nerves.Package.Providers.Docker, []}]

|nerves_bootstrap| Loadpaths Start

|nerves_bootstrap| Env Start

|nerves_bootstrap| Env End

NERVES_SYSTEM is unset
```
### Current behavior

I forked the rpi0 nerves system from Tim [(tmecklem/nerves_system_rpi0),](https://github.com/tmecklem/nerves_system_rpi0) cloned it and did a `mix deps.get`. The following `mix` job failed: 

```
 mix
|Elixir.Nerves.Package.Providers.HTTP| Downloading Artifact nerves_system_rpi0-0.12.0-dev.armv6_rpi_linux_musl.tar.gz
  From Location:
    https://github.com/tmecklem/nerves_system_rpi0/releases/download/v0.12.0-dev/nerves_system_rpi0-v0.12.0-dev.tar.gz

|Elixir.Nerves.Package.Providers.HTTP| No Available Locations

|Elixir.Nerves.Package.Providers.Docker| Creating Docker container -KbCw3XbD-X-sYJ5-nerves_system_rpi0-FA897B0EEB99FDD29C9711F868394CEACD25D190C643CFE4E419B0806C9F8236

[...]

PR #111, opened 2017-04-14
Invalid docker id

alfert:
Fixes bug #110


mobileoverlord:
We should handle this better. In this situation, you are getting a leading `-` but we should limit the character set that is being used to randomly generate the project id.

mobileoverlord:
@alfert thanks for submitting a pr, but I fixed this in a different PR by limiting the characters that could be used to generate the container id to only alpha numeric and `_`

https://github.com/nerves-project/nerves/pull/116

Thanks for reporting :bowtie:

PR #116, opened 2017-04-22
generate random alpha num for docker id

mobileoverlord:
fixes https://github.com/nerves-project/nerves/issues/110


tmecklem:
@mobileoverlord, should it be a concern that repeated calls to `Nerves.Utils.random_alpha_num(16)` return the same result for me?

tmecklem:
This looks 👍 to me.

Issue #112, opened 2017-04-14
Dockerfile fails running locale-gen stage: /bin/sh: 1: locale-gen: not found


cthree:
The Dockerfile template fails to build:
```
erik@sparrow:nerves_system_rpi0 (master=) $ mix 
==> distillery
Compiling 18 files (.ex)
Generated distillery app
==> nerves
Compiling 25 files (.ex)
Generated nerves app
==> nerves_system_br
Generated nerves_system_br app
==> nerves_toolchain_ctng
Compiling 2 files (.ex)
Generated nerves_toolchain_ctng app
==> nerves_toolchain_armv6_rpi_linux_gnueabi
Compiling 1 file (.ex)
Generated nerves_toolchain_armv6_rpi_linux_gnueabi app
==> nerves_system_rpi0
Generated nerves_system_rpi0 app
|Elixir.Nerves.Package.Providers.HTTP| Downloading Artifact nerves_system_rpi0-0.12.0-dev.armv6_rpi_linux_musl.tar.gz
  From Location:
    https://github.com/tmecklem/nerves_system_rpi0/releases/download/v0.12.0-dev/nerves_system_rpi0-v0.12.0-dev.tar.gz

|Elixir.Nerves.Package.Providers.HTTP| No Available Locations

|Elixir.Nerves.Package.Providers.Docker| Create Image

The Nerves Docker provider needs to create the image.
Proceed?  [Yn] 
Sending build context to Docker daemon 3.584 kB
Step 1/10 : FROM ubuntu:xenial
 ---> 6a2f32de169d
Step 2/10 : MAINTAINER Justin Schneck
 ---> Using cache
 ---> 9d5dadcbd004
Step 3/10 : ENV NERVES_BR_DL_DIR /nerves/cache/buildroot
 ---> Using cache
 ---> fe23d0f005e0
Step 4/10 : ENV LANG en_US.UTF-8
 ---> Using cache
 ---> 2795082da495
Step 5/10 : ENV LANGUAGE en_US:en
 ---> Using cache
 ---> f95777e4edb0
Step 6/10 : ENV LC_ALL en_US.UTF-8
 ---> Using cache
 ---> b86b9624a355
Step 7/10 : RUN locale-gen en_US.UTF-8
 ---> Running in 5c45f6ee4b58
/bin/sh: 1: locale-gen: not found
The command '/bin/sh -c locale-gen en_US.UTF-8' returned a non-zero code: 127
[...]

PR #117, opened 2017-04-22
fix locale issues in docker image

mobileoverlord:
fixes https://github.com/nerves-project/nerves/issues/112


tmecklem:
I get this when I build with these changes:

```
...
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = "en_US:en",
    LC_ALL = "en_US.UTF-8",
    LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
...
```

It looks like an error when it scrolls past on the screen. The build continues and `RUN localedef -i en_US -f UTF-8 en_US.UTF-8` seems to do fine afterward. Despite that confusing hiccup, this seems to be working fine as a fix for the problem.

I 👍 this.

PR #119, opened 2017-05-03
split lock files by target

mobileoverlord:
Splitting lock files by target will eliminate cross target dependency issues.

### DIFF
--- a/bootstrap/templates/new/mix.exs
+++ b/bootstrap/templates/new/mix.exs
@@ -20,7 +20,8 @@ defmodule <%= app_module %>.Mixfile do
      config_path: "../../config/config.exs",
      lockfile: "../../mix.lock",<% else %>
      deps_path: "deps/#{@target}",
-     build_path: "_build/#{@target}",<% end %>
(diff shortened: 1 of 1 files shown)


bootstrap/templates/new/mix.exs
```diff
@@ -20,7 +20,8 @@ defmodule <%= app_module %>.Mixfile do
      config_path: "../../config/config.exs",
      lockfile: "../../mix.lock",<% else %>
```
GregMefford: I think this should be `"../../mix.lock.#{@target}"` as well, if we're going to continue to support ☔️  instead of just :poncho: projects. :trollface:

mobileoverlord: You know, I thought about this. And I was curious if this was the case, or actually a contribution to why I'm really wasn't working. I guess you're right, but in the sense of umbrellas there is a lot of assumption of what can be run at the top of it, where targets would blow other projects minds

mobileoverlord:
@GregMefford Updated this to add target to umbrellas

PR #121, opened 2017-05-03
include the system in the runtime

mobileoverlord:
Systems contain configuration and therefore they can be interesting to include as part of the runtime.


GregMefford:
We should document what things need to come in at runtime from the System as opposed to `nerves_runtime`. I'm hesitant to bring these back in as runtime dependencies if we don't need to, just because of unnecessary namespace clutter they'll bring along.

mobileoverlord:
Closing this considering we don't have any valid use case yet.

PR #122, opened 2017-05-09
Update nerves.new docs

fhunleth:
The two commits update the nerves.new help and the nerves.new information after a successful run to be more similar to 'mix new' and add more details.

### DIFF
--- a/bootstrap/lib/mix/tasks/nerves/new.ex
+++ b/bootstrap/lib/mix/tasks/nerves/new.ex
@@ -150,24 +150,31 @@ defmodule Mix.Tasks.Nerves.New do
   defp print_mix_info(path, extra) do
     command = ["$ cd #{path}"] ++ extra
     Mix.shell.info """
-    All set!
-      #{Enum.join(command, "\n")}
+    Your Nerves project was created successfully.
 
-    Next, pick a deployment target.
-      For example: `rpi3` for Raspberry Pi 3
-      More info on targets: https://hexdocs.pm/nerves/targets.html#content
+    You should now pick a target. See https://hexdocs.pm/nerves/targets.html#content
+    for supported targets. If your target is on the list, set `MIX_TARGET`
+    to its tag name:
 
-    To set the target you can either
+    For example, for the Raspberry Pi 3 you can either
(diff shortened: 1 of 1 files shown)


bootstrap/lib/mix/tasks/nerves/new.ex
```diff
@@ -143,24 +150,31 @@ defmodule Mix.Tasks.Nerves.New do
...
-    To set the target you can either
+    For example, for the Raspberry Pi 3 you can either
       $ export MIX_TARGET=rpi3
-    Or prefix your commands
+    Or prefix `mix` commands like the following:
       $ MIX_TARGET=rpi3 mix firmware
 
-    Finally, Create firmware
+    If you are using an unsupported target or a custom system image,
+    update the `mix.exs` dependencies to point to desired system image.
```
GregMefford: Should this just be "system" instead of "system image"?

bootstrap/lib/mix/tasks/nerves/new.ex
```diff
@@ -143,24 +150,31 @@ defmodule Mix.Tasks.Nerves.New do
...
+
+    Now download the dependencies and build a firmware image:
```
GregMefford: I think this should just be "firmware" rather than "firmware image", since it's technically a firmware archive (`.fw`) unless you tell it to build an actual SD card image (`.img`).

bootstrap/lib/mix/tasks/nerves/new.ex
```diff
@@ -143,24 +150,31 @@ defmodule Mix.Tasks.Nerves.New do
...
 
-    To set the target you can either
+    For example, for the Raspberry Pi 3 you can either
       $ export MIX_TARGET=rpi3
-    Or prefix your commands
+    Or prefix `mix` commands like the following:
       $ MIX_TARGET=rpi3 mix firmware
 
-    Finally, Create firmware
+    If you are using an unsupported target or a custom system image,
```
GregMefford: The phrase "unsupported target" seems redundant to me, given that a target not supported by a standard system would require them to use a custom system.

fhunleth:
Addressed all comments. Thanks!

PR #123, opened 2017-05-12
Add commented-out instructions for enabling rootfs_additions

fhunleth:
While this information is in the Nerves instructions, having it here as
well is really convenient and saves having to look it up.


GregMefford:
👍 Thanks!

Issue #86, opened 2017-03-04
Discuss erlang.cookie error when trying to write to read only FS


mobileoverlord:
On the FAQ's we should add a blurb discussing what the obscure error received when Erlang tries to write an erlang.cookie file on the RO FS.

**How to reproduce**
unset `-name` and `-sname` and `-setcookie` in vm.args
Then, boot into the vm and try to start distributed erlang.
```
:os.cmd 'epmd -daemon'
:net_kernel.start([:"my_app@0.0.0.0"])
```

**Solution**
Ensure that a cookie is set in the vm.args even if the vm is starting in no distribution.

GregMefford:
We should also use this as an opportunity to warn that the `-extra` option consumes the rest of the options in an unintuitive way, as discovered by @electricshaman.

For example, the following won't work:

```
## Start the Elixir shell
-noshell
-user Elixir.IEx.CLI
-extra --no-halt
-name dev@192.168.0.17
-setcookie somethingsecret
```

Instead, it has to be:
```
## Start the Elixir shell
-noshell
-user Elixir.IEx.CLI
-name dev@192.168.0.17
-setcookie somethingsecret
-extra --no-halt
```

I'm not sure whether this behavior of `-extra` is a bug or by design.

fhunleth:
I was hoping that this commit to the new project template would help prevent argument ordering mistake with `-extra`: https://github.com/nerves-project/nerves/commit/a3d09db9c9623d54918371b19b9a5cd3d674865e

Maybe it needs to be more obnoxious like "whatever you do, don't add anything after this" :)

That change may not have been released yet, so maybe @electricshaman didn't have the comment in his vm.args.

mobileoverlord:
We can add this to Nerves bootstrap 0.3.1

electricshaman:
@fhunleth Definitely didn't see the comment!

GregMefford:
To be clear, @electricshaman figured out what the problem was, but it was someone in the Slack channel who actually ran into the problem "in real life." I agree that a prominent comment in the template would be a good idea.

@fhunleth I think the comment in the PR you linked above isn't clear unless you already know about the problem, so it's a good reminder to you but not obvious to someone else. When it says "add custom options here," it's easy to mistakenly read that as "here are some custom options; you should add yours to the end of the list."

I'd propose something at this level of obnoxious, to make it dead simple to do the right thing:

```

## Start the Elixir shell
-noshell
-user Elixir.IEx.CLI

[...]

PR #127, opened 2017-05-26
breakout targets in mix file

mobileoverlord:
Added new proj gen tests.
Break out targets in the mix file
Added a common place for adding deps to all targets
aligns style with https://github.com/nerves-project/nerves-examples/pull/54
resolves https://github.com/nerves-project/nerves/issues/109

PR #128, opened 2017-05-26
Set the cookie when generating new projects

mobileoverlord:
If the cookie is not set when the vm boots, and distribution is started manually, an error is raised when the system attempts to create a cookie file on the read only filesystem.

fixes https://github.com/nerves-project/nerves/issues/86


mobileoverlord:
This needs to be rebased after https://github.com/nerves-project/nerves/pull/127 is merged

bootstrap/test/nerves_new_test.exs
```diff
@@ -59,4 +59,13 @@ defmodule Nerves.BootstrapTest do
     end
   end
 
+  test "new project cookie set", context do
```
GregMefford: It would be nice to have a test that it generates a 64-character random cookie if you don't specify the option.

bootstrap/lib/mix/tasks/nerves/new.ex
```diff
@@ -70,7 +70,7 @@ defmodule Mix.Tasks.Nerves.New do
       mix nerves.new blinky --module Blinky
   """
```
GregMefford: Add docs for the `--cookie` option.

bootstrap/templates/new/rel/vm.args
```diff
@@ -1,6 +1,14 @@
...
+##  for read only filesystem.
+
+# -name <%= app_name %>@0.0.0.0
+-setcookie <%= cookie %>
+
 ## Start the Elixir shell
+
 -noshell
 -user Elixir.IEx.CLI
 -extra --no-halt
```
GregMefford: In order to fully satisfy #86, I think we should move the comment above the `-extra` option to make it even more obvious that the ordering of the options here is significant, and maybe re-word it like this:

```elixir
## Options added after -extra are interpreted as plain arguments and
## can be retrieved using :init.get_plain_arguments().
-extra --no-halt
```

bootstrap/test/nerves_new_test.exs
```diff
@@ -68,4 +68,13 @@ defmodule Nerves.BootstrapTest do
     end
   end
 
+  test "new project default cookie set", context do
+    in_tmp context.test, fn ->
+      Mix.Tasks.Nerves.New.run([@app_name])
+      assert_file "#{@app_name}/rel/vm.args", fn file ->
+        assert file =~ ~r/.+?(?:-setcookie ).{64}\n/s
```
GregMefford: If this works, I'm OK with it, but I think you could simplify a few things:
* I think `.+?` is probably the same as `.*`, but I've never seen `+?` used together like that so maybe it does something else.
* I think the `(?:` ... `)` structure is used to specify a non-capturing match group, which is doesn't seem like something you need here.
* It would be more reliable to anchor on `$` as the end-of-line instead of `\n`. For example, on Windows, it'll be `\r\n`, so there will be 65 characters there instead of `64`, but `$` handles that for you.

mobileoverlord: looks like \w wont work because it doesn't contain all base64 chars like /

GregMefford: Ah, right. I was thinking the Base64 charset was `[a-zA-Z0-9]`, but it turns out that's only 62 characters. 😅 
So you could do `[a-zA-Z0-9+/]{64}` or `\S{64}` (for 64 non-whitespace characters).

PR #129, opened 2017-06-08
Error out when package version or type are missing

fhunleth:
I ran into this based on a typo. Previously, the message was printed and
processing carried on. If you were lucky, it would crash, but on
something later. This exits `mix` as soon as it's detected and provides
some info on where to look for what's probably a typo.

### DIFF
--- a/lib/nerves/package.ex
+++ b/lib/nerves/package.ex
@@ -119,9 +119,18 @@ defmodule Nerves.Package do
     load_nerves_config(path)
     config = Application.get_env(app, :nerves_env)
     version = config[:version]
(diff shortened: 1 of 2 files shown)


mobileoverlord:
We should probably rework the tests and remove the legacy support, which is why it did not error out when the version was not passed. This was held this way during the big transition from independent compilers to the `nerves_package` compiler. The package config files did not contain the package version pre 0.3.0

fhunleth:
That's what I get for not running the tests before clicking the PR button.

mobileoverlord:
Looks good. Thanks

PR #131, opened 2017-06-08
Error out if NERVES_SYSTEM is set incorrectly

fhunleth:
This errors out immediately when the path is wrong rather than let
execution continue to hit some other issue.


mobileoverlord:
Looks good, thanks

PR #132, opened 2017-06-15
Use `$` for all command examples

srenatus:


GregMefford:
Thank you for taking the time to improve the documentation! 🎈 🚀 
For the Nerves Project repositories and documentation, we'd like to establish consistency in our command-line and code examples using the following style guide.

We'd love to have your help in finding and correcting any inconsistencies you find in the repositories in the [nerves project organization](https://github.com/nerves-project)!

1. For step-by-step instructions, we want them to be very user-friendly, even for people who are not familiar with using a command-line interface. Therefore, we will not show the `$` or `#` prompt before the commands to be entered to make them easier to copy and paste into a terminal. For example:

    ```
    export MIX_TARGET=rpi3
    mix firmware
    mix firmware.burn
    ```

2. For showing what we expect a person to see or do on the command-line, we will include an appropriate prompt before commands that they would run, to help them see which are input and which are output. We also allow and encourage the use of small formatting changes, where possible and not confusing, to emphasize or set apart the input from the output. We also encourage the use of either three dots or an ellipsis to truncate long output or code examples, only showing the relevant part. Where possible, show a truncation in a code example using a comment rather than invalid code. For example:

    ```
    $ iex -S mix
   Erlang/OTP 19 [erts-8.2] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
    
    Compiling 6 files (.ex)
    Generated example app
    Interactive Elixir (1.4.2) - press Ctrl+C to exit (type h() ENTER for help)
    
    iex(1)> System.cmd("date", [])
    {"Fri Jun 16 09:52:12 EDT 2017\n", 0}
    
    iex(2)>
    ```

    > NOTE: I have artificially inserted blank lines before the `iex>` prompts to offset them
    
    or

    ```elixir
    # custom_rpi3/mix.exs
    
    defmodule CustomRpi3.Mixfile do
    
      # ...
    
      def project do
       [app: :custom_rpi3,
        version: @version,
        # ...
        ]
      end

      # ...
    
    end
    ```

PR #134, opened 2017-06-16
Remove checksum from docker container name

mobileoverlord:
This will fix a few things with the docker provider. Instead of  adding the checksum to the container name, the same container will be used for the life cycle of the application. It also adds the functionality to clean the artifacts, starting with the docker provider. You can call `mix nerves.clean` and if a container has been made, it will remove it. This process could be hooked into `mix clean` but we might want to play around with it a little before making it happen always on ever clean.

### DIFF
--- a/bootstrap/lib/mix/tasks/nerves/env.ex
+++ b/bootstrap/lib/mix/tasks/nerves/env.ex
@@ -12,15 +12,7 @@ defmodule Mix.Tasks.Nerves.Env do
     end
     # Env moved to :nerves, try to start it otherwise, compile
     #  :nerves_system and call initialize
-    try do
-      Nerves.Env.start()
-    rescue
-      UndefinedFunctionError ->
-        unless Code.ensure_compiled?(Nerves.Env) do
-          Mix.Tasks.Deps.Compile.run ["nerves_system", "--include-children"]
-        end
-        Nerves.Env.initialize()
-    end
+    Nerves.Env.start()
     debug_info "Env End"
     if opts[:info], do: print_env()
   end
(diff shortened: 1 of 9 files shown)


lib/nerves/package/providers/docker.ex
```diff
@@ -389,15 +396,53 @@ defmodule Nerves.Package.Providers.Docker do
...
 
+  defp container_delete(nil), do: :noop
+  defp container_delete(name) do
+    container_stop(name)
+    cmd = "docker"
+    args = ["rm", name]
+    case System.cmd(cmd, args, stderr_to_stdout: true) do
+      {_, 0} ->
+        :ok
+      {<<"Error response from daemon: ", response :: binary>>, _} ->
```
electricshaman: Part of me is worried these messages will change and break our stuff, but I don't know of a better way to handle this.  There's similarity measures `String.jaro_distance` but I don't feel great about that either.

```elixir
iex(13)> String.jaro_distance("Error response from daemon:", "Error response from daemon was:")
0.956989247311828
```

We'll just have to keep an eye out for this.

bootstrap/lib/mix/tasks/nerves/env.ex
```diff
@@ -12,15 +12,7 @@ defmodule Mix.Tasks.Nerves.Env do
     end
     # Env moved to :nerves, try to start it otherwise, compile
     #  :nerves_system and call initialize
```
mobileoverlord: 👍 agreed

mobileoverlord:
Thanks for the great feedback. I updated the PR with some of your recommendations. ❤️

PR #138, opened 2017-06-26
strip sem ver major, minor, patch before parsing it

mobileoverlord:
Docker has a sem ver version where the minor number contains a leading zero. This is causing `Version.parse/1` to return an `:error` because it does not handle leading zeros.

```
iex(1)> Version.parse("17.0.1")
{:ok, #Version<17.0.1>}
iex(2)> Version.parse("17.03.1")
:error
```

This PR implements a function to strip the leading zeros and compare.

### DIFF
--- a/lib/nerves/package/providers/docker.ex
+++ b/lib/nerves/package/providers/docker.ex
@@ -258,7 +258,7 @@ defmodule Nerves.Package.Providers.Docker do
       {result, 0} ->
         <<"Docker version ", vsn :: binary>> = result
         {:ok, requirement} = Version.parse_requirement(@version)
-        {:ok, vsn} = strip_vsn(vsn)
+        {:ok, vsn} = parse_docker_version(vsn)
         unless Version.match?(vsn, requirement) do
           error_invalid_version(vsn)
         end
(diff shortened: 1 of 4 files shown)


GregMefford:
I'm on the fence about whether to throw a regex at this instead, but I'm willing to go with at as-is if you're OK with it.

mobileoverlord:
I think it would clean things up and be an improvement if you want to take a pass at it.

electricshaman:
I can't find any tests.  You must have forgotten to commit them! :trollface:

mobileoverlord:
I didn't commit them because I knew @GregMefford wanted a chance to make regex changes. :trollface:

GregMefford:
This thread is trolling so hard right now. 🚎 :trollface: 
I'll take a pass at a regex when I get a minute.

electricshaman:
The function names are interesting.  Your `strip_int` function takes a string and returns an int. 😂

lib/nerves/package/providers/docker.ex
```diff
@@ -457,6 +456,28 @@ defmodule Nerves.Package.Providers.Docker do
...
+      Enum.map([major, minor, patch], &strip_int/1)
+    vsn =
+      "#{major}.#{minor}.#{patch}" <>
+        if pre, do: "-#{pre}", else: ""
+    Version.parse(vsn)
+  end
+
+  defp strip_int(int) do
+    {int, _} = Integer.parse(int)
+    int
```
electricshaman:
```elixir
Integer.parse(int) |> elem(0)
```

electricshaman:
Would this work?  

```elixir
iex(77)> Regex.replace(~r/(\.|^)0+(?=\d)/, "17.0.1", "\\1") |> Version.parse 
{:ok, #Version<17.0.1>}

iex(78)> Regex.replace(~r/(\.|^)0+(?=\d)/, "17.03.1", "\\1") |> Version.parse
{:ok, #Version<17.3.1>}

iex(79)> Regex.replace(~r/(\.|^)0+(?=\d)/, "17.03.1-ce", "\\1") |> Version.parse
{:ok, #Version<17.3.1-ce>}

iex(80)> Regex.replace(~r/(\.|^)0+(?=\d)/, "0017000.00300.00001000-ce", "\\1") |> Version.parse
{:ok, #Version<17000.300.1000-ce>}
```

GregMefford:
I'm good with @electricshaman's implementation above.
I'd recommend wrapping that and the part to split on `,` into a private function called `parse_docker_version` or something, so it's obvious that it's all for parsing that gnarly version format.

mobileoverlord:
@electricshaman ^^ Thanks for that beautiful code :)

PR #145, opened 2017-07-05
Tweak grammar in mix nerves.clean docs

GregMefford:


bootstrap/lib/mix/tasks/nerves/clean.ex
```diff
@@ -5,11 +5,11 @@ defmodule Mix.Tasks.Nerves.Clean do
   @shortdoc "Clean artifacts for a Nerves package"
 
   @moduledoc """
-  Cleans the build artifacts and fetch sources from given dependencies
+  Cleans the build artifacts and fetches sources for given dependencies
```
mobileoverlord: This should probably read "fetched sources"

GregMefford: Oh, hah. I had thought that was saying that it cleans the artifacts and [re-]fetches the sources, but now that I think about it, that doesn't make sense as a step in the `clean` process.

PR #146, opened 2017-07-05
Allow the nerves_package compiler to safely be invoked on an individual package

mobileoverlord:
This fixes an issue where nerves.precompile could not be called when the mix project on the top of the stack is the nerves_system_*. This will also safely check for a nerves package config before trying to load it when calling `ensure_loaded/2`

### DIFF
--- a/lib/nerves/env.ex
+++ b/lib/nerves/env.ex
@@ -88,16 +88,12 @@ defmodule Nerves.Env do
     {:ok, Nerves.Package.t} | {:error, term}
   def ensure_loaded(app, path \\ nil) do
     path = path || File.cwd!
-    config? =
-      Package.config_path(path)
-      |> File.exists?
-    if config? do
+    if File.exists?(Package.config_path(path)) do
       packages = Agent.get(__MODULE__, &(&1))
       case Enum.find(packages, & &1.app == app) do
         nil ->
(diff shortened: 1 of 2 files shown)


lib/nerves/env.ex
```diff
@@ -88,19 +88,27 @@ defmodule Nerves.Env do
...
+    config? =
+      Package.config_path(path)
+      |> File.exists?
+    if config? do
+      packages = Agent.get(__MODULE__, &(&1))
+      case Enum.find(packages, & &1.app == app) do
+        nil ->
+          case Package.load_config({app, path}) do
+            %Package{} = package ->
+              IO.inspect package
```
GregMefford: This looks like it might be debug output that should be removed.

lib/nerves/env.ex
```diff
@@ -88,19 +88,27 @@ defmodule Nerves.Env do
...
-            end)
-            {:ok, package}
-          error ->
-            error
-        end
-      package -> {:ok, package}
+    config? =
+      Package.config_path(path)
+      |> File.exists?
+    if config? do
```
GregMefford: I think it would be simpler to just do `if File.exists?(Package.config_path(path)) do`, since you're not using the `config?` variable anywhere else.

PR #150, opened 2017-07-17
Suggest `config/rootfs_additions` instead of `rootfs-additions`

GregMefford:


fhunleth:
Why put the overlay under the `config` directory? The additions seem top level.

GregMefford:
My intent here was to make the FAQ match what we generate in `mix nerves.new`: https://github.com/nerves-project/nerves/blob/master/bootstrap/templates/new/config/config.exs

I'm not sure which way is better, but I've seen it done both ways in different examples. It seems that the official systems have it at the top level, so I'm OK changing the recommendation to that if we agree it's better that way.

fhunleth:
My vote is top level in the template. I agree that we need to make a pass to make sure that at least everything that we do is consistent since there's a mix in existing projects.

GregMefford:
I think one reason we typically put them in `config/` is that, when you have different overlays for different targets (e.g. `rpi3` and `x86_64`), which one you're using would likely depend on a config file (e.g. `config/rpi3/rootfs_additions` vs. `config/x86_64_rootfs_additions`). These could still be moved to the top level, though, as either `rpi3/rootfs_additions/` or `rootfs_additions_rpi3/`. I'm on board with making these top-level in the next pass.

Also, so we don't forget the discussion we had in Slack, Buildroot's recommended name for this is `rootfs_overlay`, so we should consider moving to that convention for better consistency with Buildroot documentation (still continuing to support the `rootfs_additions` option as well, with a deprecation warning).

fhunleth:
I don't think that we can do this, but for the multiple target case, it would be nice to have a base `rootfs_overlay` and then a `rpi3/rootfs_overlay`, etc. with the board-specific changes. Not a big deal to me, though.

fhunleth:
FYI - Here's an initial step at the renaming in the rpi0 repo: https://github.com/nerves-project/nerves_system_rpi0/pull/18

mobileoverlord:
@fhunleth I agree that allowing for multiple `rootfs_overlay` paths should be able to be specified. I believe that there are two ways to approach this.

This is difficult to do because `nerves` expects a single rootfs_overlay path to be declared in the config

```
config :nerves, :firmware,
  rootfs_overlay: "rootfs_overlay"
```

We could allow a list here, but it doesn't buy us much because mix would just replace the value instead of composing them. There are a few ways to tackle this.

1. Allow dependencies to declare rootfs_overlays. 

When creating the fw, we could look through the config for dependencies for rootfs_overlay paths and compose them together.

For Example:
```elixir
# config.exs
config :nerves, :firmware,
  rootfs_overlay: "rootfs_overlay"

# rpi3.exs
config :nerves_system_rpi3, :firmware,
  rootfs_overlay: "rootfs_overlay"
```

2. Allow Mix.Config to compose the config properly. 

If we use a top level key, instead of putting it under :firmware mix could compose the sub labels together. The problem here is that labels could be arbitrarily overwritten by subsequent configs and that kind of error could not be caught.

For Example:
```elixir
# config.exs
config :nerves, :rootfs_overlay,
  main: "rootfs_overlay"

# rpi3.exs
config :nerves, :rootfs_overlay,
  rpi3: "rpi3/rootfs_overlay"
```

3. Accumulate paths in an Agent.

Nerves.Env will contain an Agent with information about the build env for all Nerves.Packages. We could use this as a location for storing the information about the rootfs_overlays.

For Example:
```elixir
# config.exs
Nerves.Env.put_rootfs_overlay("rootfs_overlay")

# rpi3.exs
Nerves.Env.put_rootfs_overlay("rpi3/rootfs_overlay")
```

fhunleth:
I'm not seeing why replacing the value (option 0 - support a list) is that bad. If you split out the config into multiple files, they're still tightly coupled and you'll be looking at both the common file and the target-specific simultaneously for each change. At least, I think that you'd be mentally merging the configs in your head to make sure that the overall version is what you want.

I'm also not even sure that making a file per target is even the best way to handle target deltas since it doesn't feel like there's that much difference between target configs. For example, another way of writing the config would be to do the substitution in the main `config.exs` like this:

```elixir
config :nerves, :firmware,
  rootfs_overlay: ["common_overlay", "#{Mix.Project.config[:target]}_overlay"]
```
I guess that I need more examples since I don't run into multiple target-specific customizations in my projects.

GregMefford:
I think what Frank is proposing might be the simplest and least-surprising thing to do, since it's very similar to what we have currently and doesn't involve any "magic" to decide how to merge multiple configuration settings.

PR #151, opened 2017-07-17
Correct code_reloader parameter

tokafish:
I'm not sure if phoenix renamed this parameter at some point (See https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/config/dev.exs#L25), but now it's `code_reloader`. Incorrectly naming the parameter in this example causes Phoenix to crash, since live_reload isn't available on the device.

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -62,7 +62,7 @@ config :ui, Ui.Endpoint,
   server: true,
   render_errors: [accepts: ~w(html json)],
   pubsub: [name: Nerves.PubSub],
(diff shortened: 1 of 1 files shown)


GregMefford:
You're right! Thanks for taking the time to improve the documentation for others! 🎈 🚀

GregMefford:
I was going to go fix it in the example code too but found that it was already correct. Must have just been a typo in the docs!
https://github.com/nerves-project/nerves-examples/blob/master/hello_phoenix/apps/ui/config/dev.exs#L12

PR #155, opened 2017-07-21
Add `mix nerves.system.shell` to `nerves_bootstrap`

GregMefford:
- [x] Check that `pkg` is a system dep
- [x] Check that `docker` is installed
- [x] Do something equivalent on Linux

This allows you to run `mix nerves.system.shell` from either a project directory or a custom system source directory. From a project directory, it will put you in a system-configuration shell for the project's current target system. From a system source directory, it will put you in a shell for that system, obviously.

In order to try it out, pull down the `nerves` repo, then do:

```bash
cd bootstrap
mix install
```

> Note: when you're done testing, use `mix local.nerves` to revert to the latest official release.

Then, create a test project and add in `nerves` as a `path` dependency with `override: true`.
You should be able to follow the instructions here to create a custom system for testing: https://hexdocs.pm/nerves/systems.html#customizing-your-own-nerves-system.

When you create the new custom system according to the docs, try running `mix nerves.system.shell` before you've actually compiled the system for the first time. It should allow you to get to the shell and do things like `make menuconfig` and `make savedefconfig` without ever having compiled it. Then, you can `exit` or `CTRL+D` out of the system config shell and `mix firmware` to have it build the system and the firmware. Unfortunately, the `mix nerves.info` task still tries to build the whole system before telling you about the environment, but that's hard to fix because it's in the `nerves` dependency instead of the `nerves_bootstrap` archive. I decided to shave that yak separately. Other than that, and the current inability to `CTRL+C` while in the system config shell, I think everything should work the way you expect it to.

One small but potentially-significant tweak I made to the Docker provider is to move where it stores the container ID, used to associate a system with its build container:

```diff
-  defp container_id_file() do
-    Mix.Project.build_path
-    |> Path.join("nerves/.docker_id")
+  defp container_id_file(pkg) do
+    Artifact.base_dir(pkg)
+    |> Path.join(".docker_id")
   end
```

This would cause existing containers to be invalidated and re-created (causing a full re-build), but then allows the same system to be used on multiple projects and share a container, because the ID is part of the system source instead of being different for each project that uses this system. This is already where the compiled artifacts were being stored, so it makes more sense this way.

I currently **didn't implement the similar functionality on Linux yet**, but my intention was that it would mimic the same behavior by putting you into a sub-shell in the appropriate directory for the system, with appropriate environment variables set. I hate to push anyone's buttons, but we could also **just use Docker on Linux**, too. It would not really hurt performance noticeably and would simplify the installation instructions since all the required dependencies would already be in the container. I'm definitely **looking for feedback and discussion** on this.

I'm also considering setting the default prompt in the system config shell to something like `root@nerves_system_shell:/nerves/build/#` or maybe just `nerves system> ` to make it more obvious that this isn't your normal user shell environment and you should `exit` out of it when you're done configuring the system.


fhunleth:
This sounds like a great improvement on Mac! I guess that my feelings are that there's a high bar to reach before this is going to be acceptable on Linux, since the situation on Linux is currently pretty good (possibly barring documentation, though). 

Is there any way to get CTRL-C to work in the shell, though? I realize that this is one of many Erlang annoyances that we run into when invoking shell applications, but it really has a way of killing things. For example, the Nerves.Runtime.Shell is super cool, but between CTRL-C not working in it and remsh not working, a lot of people don't use it and it's really sad.

GregMefford:
I'm still hopeful that we could somehow get around the CTRL+C issue, but I didn't want to hold up the initial release of this capability because it's such a big step forward even with that problem being unsolved. I think fixing the CTRL+C problem would likely involve patches to Erlang, Elixir, and/or Mix.

mobileoverlord:
This looks great so far. I think we should default to linux opening a sub shell and setting the cwd to the output directory.

electricshaman:
While researching some things in response to @GregMefford's comments about using Docker in Linux, I came across this study by IBM: [An Updated Performance Comparison of Virtual Machines
and Linux Containers](http://domino.research.ibm.com/library/cyberdig.nsf/papers/0929052195DD819C85257D2300681E7B/$File/rc25482.pdf)

I haven't fully settled on an _informed_ opinion yet, but I felt the study was relevant enough to share.

GregMefford:
Interesting paper, thanks for sharing, @electricshaman.

Surprising result in the abstract (which is all I've read so far):
> Our results show that containers result in equal or better
> performance than VMs in almost all cases. Both VMs and
> containers require tuning to support I/O-intensive applications.

GregMefford:
I think this is ready to go at this point, unless someone has additional feedback.
The prompt is intentionally differentiated from your normal system prompt both on OSX and Linux, so you remember that you're in a sub-shell.

Does anyone have thoughts on whether/how we should shorten the prompt on Linux?

OSX:
screen shot 2017-07-26 at 11 13 16 pm

Linux:
screen shot 2017-07-26 at 11 13 41 pm

electricshaman:
@GregMefford Personally I think it's fine the way it is.  I like knowing the full path.  It might not be the most aesthetically pleasing thing to look at, but it kind of comes with the territory.  And showing the full path is never going to cause any confusion about where the root is or how it's shortened, etc.

mobileoverlord:
This is fantastic work. Thank you 😊

Issue #152, opened 2017-07-18
Application partition documentation is out of date


fhunleth:
The current documentation talks about the FAT filesystem for persisting data. It's now EXT4. More information is at https://github.com/nerves-project/nerves_runtime#filesystem-initialization. Some of that could be reused or modified to make it more accessible to a new Nerves user.

mobileoverlord:
Looks like these docs are up to date and discuss how the default application data partition is formatted with ext4

mobileoverlord:
Sorry I was looking at the wrong docs. I reopened this issue.

GregMefford:
Fixed by #159

PR #159, opened 2017-07-31
update filesystem docs to talk about using ext4 instead of fat

mobileoverlord:
fixes https://github.com/nerves-project/nerves/issues/152


docs/Advanced Configuration.md
```diff
@@ -143,10 +143,15 @@ These settings are defined in `etc/erlinit.config`.
...
 +----------------------------+
-| p2: App Data  (FAT32)      |
+| p2: App Data  (ext4)       |
 +----------------------------+
 ```
 
+More information about how the App Data partition is initialized and mounted can be found
+in the documentation for `nerves_runtime` [Filesystem Initialization](https://hexdocs.pm/nerves_runtime/readme.html#filesystem-initialization)
+
+**Additional Partitions**
```
GregMefford: I think this should be an h3 (prefix with `###`) instead of just bold text.
It would also be better to have this section be "Adding a Partition" to fit with the "Mounting the Partition" section that follows it.

Issue #161, opened 2017-08-02
Better handling of clean vs. incremental builds for custom systems


mobileoverlord:
`mix nerves.system.shell` upon initialization calls `make clean`, which prevents introspection into the build. `make clean` should be called only by `mix nerves.clean`

GregMefford:
For posterity, the decision to do it this way was that people often don't (and potentially shouldn't) understand at first that Buildroot doesn't handle incremental builds properly, so it's easy to get into a state where the build completes "successfully" but hasn't actually built what you wanted, so it doesn't work and it's hard to figure out why.

We had talked about having the build do a `make clean` by default unless an environment variable was specified so that people familiar with Buildroot could skip that step if they wanted to. To be honest, the reason this variable isn't there now is that I forgot to go back and add it.

Here are a few paths forward as I see them:

1. As this issue suggests, never `make clean` unless the user asks for it explicitly via `make clean`. This would give the most flexibility but also requires that a newcomer to custom Nerves systems would need to know that this step is required to get a reliable build. Maybe that's fine, since they're going to have to read a guide anyway to figure out the `make menuconfig`/`make savedefconfig` dance.

2. Don't `make clean` when running `mix nerves.system.shell`, but always run it from the `mix compile`/`mix firmware` workflow. This gives you a change to go do whatever you want and have full control from within the system configuration shell (without a surprise `make clean`), but when you're done, you always get a clean build. Along with this option, we could still have a `NERVES_SYSTEM_BUILD_INCREMENTAL` variable that, if set, would cause `mix compile`/`mix firmware` to skip the `make clean` step.

electricshaman:
I am strongly in the camp of #1.

electricshaman:
Option one is also consistent with how Nerves takes a view of things when using the Local provider.  https://github.com/nerves-project/nerves/pull/148

GregMefford:
I'm on board with option 1 as well, and it's the simplest to implement (deleting one line from each provider). I'll make a PR to do that for now, since option 2 is actually still part of option 1 and could be implemented separately if desired.

fhunleth:
I'll say that I'm in the camp of consistent builds. Here's some good information to read for anyone wanting to do incremental Buildroot rebuilds: http://nightly.buildroot.org/manual.html#full-rebuild

[...]

PR #163, opened 2017-08-03
Remove `make clean` from `nerves.system.shell` and refresh System docs

GregMefford:
* `mix nerves.system.shell` does not run `make clean` automatically (see #161)
* Better error-checking and error messages in a few cases that I hadn't hit until now (like `mix deps.get` not having been run yet)
* Updated System docs to be current and also cover `nerves.system.shell`

**Specific feedback requested:**

I'm not 100% clear on whether I got the `make linux-savedefconfig` and `make busybox-savedefconfig` documentation right, since that's not something I've done very often. Let me know if it's not right. I feel like it might have been automated so you don't have to manually copy the files anymore, but I wasn't sure.


ConnorRigby:
Im not sure if it is fixed yet, but as of feb 2016, (that was the last time i tried) `make linux-savedefconfig` and `make busybox-savedefconfig` did not work for me as intended. (or i read the docs wrong)

fhunleth:
Regarding the `make *-savdefconfig` commands, the expected Buildroot behavior is to keep the resulting `defconfig` files in the `build` directories. It's wrong to copy them back to the source since you may have instructed Buildroot to concatenate several defconfig fragments together and it won't know how to split them up again. I also believe that I saw an discussion that the `make savedefconfig` implementation in Buildroot that would save back to the original `defconfig` file was not desired.

All of this makes things more manual, but since my workflow is to meticulously check the defconfig deltas whenever I copy the new defconfigs over, I'm not sure that it costs me any time.

ConnorRigby:
Ah ok. I do it all manually also. But last time i tried to use those commands (specifically busybox) - it actually failed due to a missing file and i haven't tried it since. The docs look good to me tho @GregMefford

mobileoverlord:
Looks good to me. Thanks @GregMefford

Issue #164, opened 2017-08-06
rootfs_additions -> rootfs_overlay


fhunleth:
The `rootfs_additions` config variable and directory name need to be updated to `rootfs_overlay` now that we're following the Buildroot naming convention. A `git grep rootfs_additions` looks like it finds everything that needs to change. The only trick is to make sure that the `rootfs_additions` key still works so as to not break existing projects.

GregMefford:
Resolved by #165.

PR #165, opened 2017-08-07
Replace `rootfs_additions` with `rootfs_overlay`

GregMefford:
Also changed the recommendation to put the `rootfs_overlay` directory at the project root instead of under `config`.


GregMefford:
Addresses #164.

PR #168, opened 2017-08-08
Fix hung nerves.system.shell

mobileoverlord:
Thanks @GregMefford ❤️


GregMefford:
We still need to bump versions and update the changelog. I can knock that out shortly.

PR #170, opened 2017-08-09
append system deps as a list

mobileoverlord:
System apps should be appended as a list so there is a clear place for putting deps that are only for a certain system. For example, I have a project with qemu and rpi0 and only rpi0 also needs nerves_init_gadget.

```elixir
# Specify target specific dependencies
  def deps("host"), do: []
  def deps(target) do
    [
      {:bootloader, "~> 0.1"},
      {:nerves_runtime, "~> 0.4"},
    ] ++ system(target)
  end

  #def system("qemu_arm"), do: {:nerves_system_qemu_arm, ">= 0.0.0", runtime: false}
  def system("qemu_arm"), do: [{:nerves_system_qemu_arm, path: "../../nerves_system_qemu_arm"}]
  def system("rpi0") do
    [
      {:nerves_system_rpi0, "~> 0.17.0"},
      {:nerves_init_gadget, github: "fhunleth/nerves_init_gadget"}
    ]
  end
  def system(target), do: Mix.raise "Unknown MIX_TARGET: #{target}"
```

### DIFF
--- a/bootstrap/templates/new/mix.exs
+++ b/bootstrap/templates/new/mix.exs
@@ -62,14 +62,14 @@ defmodule <%= app_module %>.Mixfile do
   # Specify target specific dependencies
   def deps("host"), do: []
   def deps(target) do
-    [ system(target),
+    [
       {:bootloader, "~> <%= bootloader_vsn %>"},
       {:nerves_runtime, "~> <%= runtime_vsn %>"}
-    ]
+    ] ++ system(target)
   end
 
 <%= for target <- targets do %>
-  def system("<%= target %>"), do: {:<%= "nerves_system_#{target}" %>, ">= 0.0.0", runtime: false}
+  def system("<%= target %>"), do: [{:<%= "nerves_system_#{target}" %>, ">= 0.0.0", runtime: false}]
 <% end %>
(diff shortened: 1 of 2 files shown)


electricshaman:
I'm not crazy about this.  It would be easy for folks to lose the pipe.  What about using List.flatten on the deps key?   

```elixir
  def project do
    [app: :squitter_firmware,
     # ...
     deps: List.flatten(deps())]
  end

  def deps do
    [{:nerves, "~> 0.7"},
     {:nerves_firmware_ssh, github: "fhunleth/nerves_firmware_ssh"},
     {:squitter_web, path: "../web"},
     {:squitter, path: "../squitter"},
     deps(@target)]
  end

  def system("rpi3") do
    [{:nerves_system_rpi3_sdr, github: "electricshaman/nerves_system_rpi3_sdr", runtime: false}]
  end
```

fhunleth:
Off-topic, but a couple projects referenced above are now on hex:

`{:nerves_firmware_ssh, "~> 0.2.0"}`
`{:nerves_init_gadget, "~> 0.1.0"}`

electricshaman:
Can ya'll hear me on this thing? ;)

PR #172, opened 2017-08-11
Nerves release plugin

mobileoverlord:
This will simplify the release config and move the settings for Nerves into a plugin.

Instead of this
```elixir
release :example_proj do
  set version: current_version(:example_proj)
  plugin Bootloader.Plugin
  if System.get_env("NERVES_SYSTEM") do
    set dev_mode: false
    set include_src: false
    set include_erts: System.get_env("ERL_LIB_DIR")
    set include_system_libs: System.get_env("ERL_SYSTEM_LIB_DIR")
    set vm_args: "rel/vm.args"
  end
end
```

we use this
```elixir
release :example_proj do
  set version: current_version(:example_proj)
  plugin Bootloader
  plugin Nerves
end
```

This PR also adds a commit to skip the archival release phase.

Depends on:
https://github.com/nerves-project/bootloader/pull/6

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -64,7 +64,7 @@ defmodule Mix.Tasks.Firmware do
   end
 
   defp build_release do
-    Mix.Task.run "release", ["--silent"]
+    Mix.Task.run "release", ["--silent", "--no-tar"]
   end
 
   defp build_firmware(system_path) do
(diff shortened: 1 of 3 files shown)


mobileoverlord:
In our use case the defaults are almost never going to need to be overridden. If the developer would want to override what the `Nerves` plugin for distillery is doing they can omit the plugin declaration and inline the code.

PR #174, opened 2017-08-14
remove legacy compiler key from package struct

mobileoverlord:
This was only needed to support finding artifacts from the `nerves_system` and `nerves_toolchain` compilers which have long been deprecated.

### DIFF
--- a/lib/nerves/package.ex
+++ b/lib/nerves/package.ex
@@ -20,7 +20,6 @@ defmodule Nerves.Package do
     config pkg, :nerves_env,
       type: :system,
       version: version,
-      compiler: :nerves_package,
       artifact_url: [
         "https://github.com/nerves-project/\#{pkg}/releases/download/v\#{version}/\#{pkg}-v\#{version}.tar.gz",
       ],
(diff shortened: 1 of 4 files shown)


lib/nerves/package.ex
```diff
@@ -133,7 +127,6 @@ defmodule Nerves.Package do
     end
     platform = config[:platform]
     provider = provider(app, type)
-    compiler = config[:compiler]
```
fhunleth: How do you feel about having a deprecation warning here to encourage everyone to remove this from their `nerves.exs` files?

mobileoverlord: Added overall deprecation here https://github.com/nerves-project/nerves/pull/176

lib/nerves/package/artifact.ex
```diff
@@ -56,7 +56,6 @@ defmodule Nerves.Package.Artifact do
     else
       base_dir(pkg)
       |> Path.join(name(pkg, toolchain))
-      |> protocol_vsn(pkg)
```
GregMefford: I'm not sure I understand why this whole function is no longer needed. It seems like it would still be useful for manipulating the `build_path` for toolchains and systems.

mobileoverlord: configuration for the build_paths for the compiler should be done in the compiler instead of package then. This only removes some code which would look in a different area for code to support the compiler packages in the long deprecated `nerves_toolchain` and `nerves_system` apps.

mobileoverlord:
I rebased this and am ready to go for another review before merging

PR #176, opened 2017-08-14
Move Nerves config from nerves.exs to mix.exs

mobileoverlord:
The config can be read in form the mix file and packages can be identified with backwards compatibility (until after 1.0).

TL;DR
Turns this:
```elixir
# nerves.exs
use Mix.Config

version =
  Path.join(__DIR__, "VERSION")
  |> File.read!
  |> String.trim

pkg = :nerves_system_rpi0

config pkg, :nerves_env,
  type: :system,
  version: version,
  compiler: :nerves_package,
  artifact_url: [
    "https://github.com/nerves-project/#{pkg}/releases/download/v#{version}/#{pkg}-v#{version}.tar.gz",
  ],
  platform: Nerves.System.BR,
  platform_config: [
    defconfig: "nerves_defconfig"
  ],
  checksum: [
    "nerves_defconfig",
    "rootfs_overlay",
    "linux-4.4.defconfig",
    "fwup.conf",
    "cmdline.txt",
    "config.txt",
    "post-createfs.sh",
    "VERSION"
  ]
```
Into this:
```elixir
# mix.exs
defmodule NervesSystemRpi0.Mixfile do
  use Mix.Project
  
  @app :nerves_system_rpi0
  @version "0.17.0"

  def project do
    [app: @app,
     version: @version,
     nerves_package: nerves_package(),
     #..
     aliases: ["deps.precompile": ["nerves.env", "deps.precompile"]]]
  end

  def nerves_package do
    [type: :system,
     artifact_url: [
       "https://github.com/nerves-project/#{@app}/releases/download/v#{@version}/#{@app}-v#{@version}.tar.gz",
     ],
     platform: Nerves.System.BR,
     platform_config: [
       defconfig: "nerves_defconfig"
     ]]
  end
#..
end

```

Packages are currently identified as being a `Nerves.Package` if they declare their config in the `nerves.exs` file in source path. This PR changes the load process to first look in `Mix.Project.config[:nerves_package]` for the config. 

This also adds a deprecation warning to any packages that are loading the `nerves.exs` file.

### DIFF
--- a/lib/nerves/package.ex
+++ b/lib/nerves/package.ex
@@ -116,25 +116,18 @@ defmodule Nerves.Package do
   """
   @spec load_config({app :: atom, path :: String.t}) :: Nerves.Package.t
   def load_config({app, path}) do
-    load_nerves_config(path)
-    config = Application.get_env(app, :nerves_env)
+    config = config(app, path)
     version = config[:version]
-    unless version do
-      Mix.shell.error "The Nerves package #{app} does not define a version.\n\n" <>
-                      "Verify that the key exists in '#{config_path(path)}'\n" <>
-                      "and that the package name is correct."
-      exit({:shutdown, 1})
-    end
-    type = config[:type]
+    type = config[:nerves_package][:type]
     unless type do
       Mix.shell.error "The Nerves package #{app} does not define a type.\n\n" <>
                       "Verify that the key exists in '#{config_path(path)}'.\n"
       exit({:shutdown, 1})
     end
-    platform = config[:platform]
+    platform = config[:nerves_package][:platform]
     provider = provider(app, type)
-    compiler = config[:compiler]
-    config = Enum.reject(config, fn({k, _v}) -> k in @required end)
+    compiler = config[:nerves_package][:compiler]
+    config = Enum.reject(config[:nerves_package], fn({k, _v}) -> k in @required end)
 
     %Package{
       app: app,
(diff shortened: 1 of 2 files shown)


mobileoverlord:
I modified the key to `nerves_package`

fhunleth:
I may have made a mistake when approving this. If no changes need to be made to existing systems, then approve. If a change needs to be made to existing systems, then I don't approve. Rereading this PR makes it sound like you need to modify the `mix.exs` of your project. That means that our documentation is wrong so this PR should also contain updates to the docs.

lib/nerves/package.ex
```diff
@@ -213,6 +206,53 @@ defmodule Nerves.Package do
...
+          Mix.Project.config
+        end)
+      end
+    nerves_package =
+      case project_config[:nerves_package] do
+        nil ->
+          # TODO: Deprecated. Clean up after 1.0
+          load_nerves_config(path)
+          config = Application.get_env(app, :nerves_env)
+          Mix.shell.error """
```
fhunleth: This shouldn't be a deprecation until we document the new way. My suggestion is to make this silent so that we don't cause every one presenting nerves projects or doing training at ElixirConf to have to update their projects right before the conference. After ElixirConf, I say that we enable the warning so that people know to update their projects.

fhunleth:
Update: I see where you pull in `nerves.exs` and print out the deprecation. How does https://github.com/nerves-project/nerves/pull/177 apply to making sure that the old way works through ElixirConf at least.

mobileoverlord:
https://github.com/nerves-project/nerves/pull/177 is unaffected if the configuration came from mix.exs or if it came from nerves.exs.

fhunleth:
That what I was afraid of. In other words, if we revert this PR, that fixes #177.

fhunleth:
Just to resolve this thread after talking with mobileoverlord, merging this PR only added a deprecation warning if you're using `nerves.exs`. Issue #177 doesn't affect the `nerves.exs` case.

PR #177, opened 2017-08-14
parse url placeholders and replace at time of use

mobileoverlord:
In the process of moving config from nerves.exs to mix.exs we need to modify the way that the url's for toolchains are evaluated. String interpolation in the mix.exs file fails because the function calls need to resolve when evaluating the mix.exs file. This changes it so that `HOST_ARCH` and `HOST_PLATFORM` are replaced at runtime with the appropriate values from the original calls.

### DIFF
--- a/lib/nerves/package/providers/http.ex
+++ b/lib/nerves/package/providers/http.ex
@@ -3,20 +3,24 @@ defmodule Nerves.Package.Providers.HTTP do
   Downloads an artifact from a remote http location.
   """
 
-  @behaviour Nerves.Package.Provider
-
   alias Nerves.Package.Artifact
   require Logger
+
+  @behaviour Nerves.Package.Provider
+  @url_placeholders %{
+    "HOST_PLATFORM" => {Nerves.Env, :host_platform, []},
+    "HOST_ARCH" => {Nerves.Env, :host_arch, []}
+  }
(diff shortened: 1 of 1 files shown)


fhunleth:
I don't understand the issue. 

Is it that this fails:

`"https://github.com/nerves-project/toolchains/releases/download/v#{version}/#{app}-#{version}.#{Nerves.Env.host_platform}-#{Nerves.Env.host_arch}.tar.xz"`

And you want to change it to this:

`"https://github.com/nerves-project/toolchains/releases/download/v#{version}/#{app}-#{version}.HOST_PLATFORM-HOST_ARCH.tar.xz"`?

I really don't like that if that's the case.

mobileoverlord:
The issue this is solving here is that `Nerves.Env` is not available at the time the mix file is being evaluated. Nothing is, other than Mix and Elixir. Therefore, we need a way to be able to make the calls later and substitute in the results. We could quote the whole thing, we could use a different syntax. I am open to other ideas. We could even copy and paste in the two functions we want access to into the mix file in all the toolchains.

mobileoverlord:
This is going to be solved by copy and pasting the small amount of helper code needed to identify the host_arch and host_platform into the toolchains mix files

PR #180, opened 2017-08-16
Docker enhancements

mobileoverlord:
Enhancements
* create named docker volumes for build assets
* containers are transient 
* updated Dockerfile to create a faster loading base image by reducing the number of layers to 1.
* attempts to pull the image from docker hub instead of always building it from a dockerfile locally


template/Dockerfile
```diff
@@ -35,4 +34,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
   libc6:i386 \
```
fhunleth: What's up with the 32-bit libraries?

nervesbot: its enabled for linkit. But, Linkit could compose its own docker image that is based off this one and adds the 32 bit libs

fhunleth: Oh yeah, I remember. Sigh.

mobileoverlord:
We should change the download cache location from a volume to a host mount.

PR #184, opened 2017-08-18
Bump fwup requirement to v0.15.x

fhunleth:
This pulls in the `meta-misc` and `meta-vcs-identifier` support that is
being used for the Nerves system regression tests. This is starting to
cause errors for people who haven't updated `fwup` in a while, so just
catch it here. Plus `fwup` is getting close to 1.0, so catching everyone
up will be useful in exposing any issues in the latest builds.


mobileoverlord:
Should we apply this back to the staged 0.7.4-dev stuff?

fhunleth:
I'm thinking so. It would avoided joelbyler's issue from the slack channel.

PR #189, opened 2017-08-22
check the cache_dir location for a tar before downloading one

mobileoverlord:
The cache dir is set by defining `NERVES_CACHE_DIR` env var or it will default to  `~/.nerves/cache` if undefined. Before the http provider attempts to fetch an artifact, it will check in the cache dir for one. If it does end up downloading an artifact, it will be downloaded to the cache_dir location and unpacked in the artifact_dir as to follow the current workflow. Does not break backwards compatibility.


lib/nerves/env.ex
```diff
@@ -53,6 +53,18 @@ defmodule Nerves.Env do
...
 
   @doc """
+  The download cache location for artifacts.
+
+  Placing an artifact tar in this location will bypass the need for it to
+  be downloaded.
+  """
+  @spec cache_dir() :: path :: String.t
+  def cache_dir do
+    (System.get_env("NERVES_CACHE_DIR") || "~/.nerves/cache")
```
fhunleth: Reminder to change to `~/.nerves/dl`

lib/nerves/env.ex
```diff
@@ -53,6 +53,18 @@ defmodule Nerves.Env do
   end
 
   @doc """
+  The download cache location for artifacts.
+
+  Placing an artifact tar in this location will bypass the need for it to
+  be downloaded.
+  """
+  @spec cache_dir() :: path :: String.t
+  def cache_dir do
```
fhunleth: Kind of thinking that this should be called `download_dir()`...

lib/nerves/package/providers/http.ex
```diff
@@ -60,29 +60,32 @@ defmodule Nerves.Package.Providers.HTTP do
...
     shell_info "Unpacking #{artifact}", """
       To Destination:
         #{destination}
     """
     tmp_path = Path.join(destination, ".tmp")
+    File.rm_rf(tmp_path)
     File.mkdir_p!(tmp_path)
     tar_file = Path.join(tmp_path, artifact)
-    File.write(tar_file, tar)
+    File.cp(file, tar_file)
```
fhunleth: Can't the `artifact` be untarred as is rather than needing to be copied to a temporary file?

lib/nerves/package/providers/http.ex
```diff
@@ -60,29 +60,32 @@ defmodule Nerves.Package.Providers.HTTP do
...
 
     {_, status} = System.cmd("tar", ["xf", artifact], cd: tmp_path)
```
fhunleth: Are you stripping off the first directory from here to the `case` and copying the contents to the destination? If so, there's the `--strip-components=1` option that I think will let you untar directly to the destination directory without all of the copies.

mobileoverlord:
I updated and pushed the changes requested

mobileoverlord:
Abandon ship!

PR #192, opened 2017-08-26
ensure dl dir exists before trying to write to it

mobileoverlord:
Ensure that the dl dir exists before trying to write to it.
Release 0.7.4


tmecklem:
I updated to use this and it created the dl directory and used the artifact properly. Not sure what's up with the failing check, but this is 👍 for me.

Issue #196, opened 2017-09-02
Update Phoenix integration docs to use ponchos


fhunleth:
The current docs use umbrellas for pulling in Phoenix. See https://hexdocs.pm/nerves/user-interfaces.html#phoenix-web-interfaces. This is no longer a best practice, so update docs to use ponchos.

PR #201, opened 2017-09-18
Update docs for Phoenix UI

rjanja:
Initially I was just going to update the docs for umbrella project layout w/ phoenix 1.3 and nerves_network, but then found #196 and learned of the poncho projects. So instead of wholly removing the umbrella project configuration, I've attempted to separate the project organization from the configuration of networking and phoenix. I've followed these steps in both project configurations with success.
* phoenix -> 1.3
* nerves_networking 0.6.0 -> nerves_network 0.3
* poncho projects recommendation

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -3,7 +3,40 @@
 ## Phoenix Web Interfaces
 
 Phoenix makes an excellent companion to Nerves applications by offering an easy-to-use, powerful framework to create user interfaces in parallel with Nerves device code.
-The easiest way to handle this is to lay out your application as an Umbrella.
+
+### Set up your application
+
+Although it's possible to lay out your application as an Umbrella project, the preferred way is to use poncho projects, i.e. a side-by-side project layout. For the reasoning behind this, please read the [Poncho Projects](http://embedded-elixir.com/post/2017-05-19-poncho-projects/) blog post by Greg Mefford.
+
+We'll cover the use of both approaches here, but we encourage you to use ponchos.
+
+#### Using a poncho project layout
+
+First, generate the two new apps in a containing folder:
+
+```bash
+$ mkdir nervy && cd nervy
+$ mix nerves.new fw
+$ mix phx.new ui --no-ecto --no-brunch
+```
+
+Now, add the Phoenix `ui` app and the `nerves_network` library to the `fw` app as dependencies:
+
+```elixir
+# fw/mix.exs
+
+# ...
+def deps do
+  [{:nerves_network, "~> 0.3"},
+   {:ui, path: "../ui"}]
+end
+# ...
+```
+
+Next: [Configure Networking](#configure-networking)
+
+
+#### Using an umbrella project layout
 
 First, generate a new umbrella app, called `nervy` in this case:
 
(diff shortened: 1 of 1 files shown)


fhunleth:
@rjanja Thanks for doing this! Your PR looks good to me. 

@GregMefford @mobileoverlord Good for both of you?

rjanja:
Great! Thanks for all your guys' work on Nerves, it's come so far in the past year! I'm happy to contribute where I can as a Nerves newbie.

docs/User Interfaces.md
```diff
@@ -3,7 +3,40 @@
 ## Phoenix Web Interfaces
 
 Phoenix makes an excellent companion to Nerves applications by offering an easy-to-use, powerful framework to create user interfaces in parallel with Nerves device code.
-The easiest way to handle this is to lay out your application as an Umbrella.
+
+### Set up your application
+
+Although it's possible to lay out your application as an Umbrella project, the preferred way is to use poncho projects, i.e. a side-by-side project layout. For the reasoning behind this, please read the [Poncho Projects](http://embedded-elixir.com/post/2017-05-19-poncho-projects/) blog post by Greg Mefford.
```
GregMefford: I think we should lower-case both "Umbrella" and "Poncho" everywhere for consistency, since I don't think either is intended to be a proper name. The first sentence is also a little awkward because of the "i.e." in the middle. Maybe something like:

> Although Nerves supports umbrella projects, the preferred project structure is to simply use separate Mix projects side-by-side with `path` dependencies between them if they're in the same source code repository. We call this a "poncho project" structure.

I also don't think you need to credit me as writing the blog post on this page. Maybe something like:

> For the reasoning behind this, please see [this blog post describing poncho projects](http://embedded-elixir.com/post/2017-05-19-poncho-projects/).

docs/User Interfaces.md
```diff
@@ -16,89 +49,109 @@ Next, create your sub-applications for Nerves and for Phoenix:
...
-...
+# ...
+```
+
+##### Specifying configuration order
+
+By default, the main config loads the application configs in an unordered way: `import_config "../apps/*/config/config.exs`. This can cause problems if the  `ui` config is applied last: we may lose overrides applied in the `fw` config. You need to force the order in which the config files get imported:
+
+```elixir
+# nervy/config/config.exs
```
GregMefford: It seems like you prefer to not mention `nervy` in the file path (which is fine with me). Let's be consistent and not include it here either.

docs/User Interfaces.md
```diff
@@ -16,89 +49,109 @@ Next, create your sub-applications for Nerves and for Phoenix:
...
-   {:nerves_networking, "~> 0.6.0"}]
+   {:nerves_network, "~> 0.3"}]
 end
-...
+# ...
+```
+
+##### Specifying configuration order
+
+By default, the main config loads the application configs in an unordered way: `import_config "../apps/*/config/config.exs`. This can cause problems if the  `ui` config is applied last: we may lose overrides applied in the `fw` config. You need to force the order in which the config files get imported:
```
GregMefford: config -> configuration

docs/User Interfaces.md
```diff
@@ -16,89 +49,109 @@ Next, create your sub-applications for Nerves and for Phoenix:
...
 
-In order to build the `ui` Phoenix application into the nerves `fw` app, you need to add some configuration to your firmware config:
+To set the default networking configuration:
 
 ```elixir
-# nervy/apps/fw/config/config.exs
+# fw/config/config.exs
+
+# ...
+# Use your ISO 3166-1 alpha-2 country code below
```
GregMefford: 😂

docs/User Interfaces.md
```diff
@@ -16,89 +49,109 @@ Next, create your sub-applications for Nerves and for Phoenix:
...
+```
+
+For more network settings, see the [`nerves_network`](https://github.com/nerves-project/nerves_network) project.
 
-use Mix.Config
 
-config :ui, Ui.Endpoint,
+### Configure Phoenix
+
+In order to build the `ui` Phoenix app into the Nerves `fw` app, you need to add some configuration to your firmware config:
```
GregMefford: app -> application

Maybe:
 > In order to build the `ui` Phoenix app into the Nerves `fw` app, you will need to make some changes to your `fw` application configuration:

docs/User Interfaces.md
rjanja: I added this in based on the likelihood that users following this guide would be adding in a `wlan0` interface and would miss this detail. But maybe it's enough to just focus on `eth0` and let wireless configuration be handled by the separate docs?

docs/User Interfaces.md
rjanja: Sure, I initially included it here because it could otherwise be confusing which `config` folder was being referenced within an umbrella project layout. But I agree it's not as consistent.

docs/User Interfaces.md
GregMefford: Ah yeah, good point. I almost suggested that you simply include the `wlan0` defaults here as well. Maybe that would be more clear, given that there's a problem where `wlan0` doesn't work right if you specify the `ipv4_address_method` option.

rjanja:
Thanks for the review and suggestions. I think I've addressed each, and have also removed mentions of "project layout" in favor of "project structure" to be more consistent for non-English speaking developers. I'm happy to make additional edits to polish this up as needed.

fhunleth:
Thanks!

PR #209, opened 2017-10-13
Contributing Documentation

GregMefford:
This PR depends on the CODE_OF_CONDUCT one, since it references that in the document.
I put both this and `CODE_OF_CONDUCT.md` in the `docs` directory so that they show up both in the Hexdocs documentation and get recognized by GitHub when opening a PR/Issue/etc.

This is still work-in-progress:

- [X] Initial draft
- [x] Figure out how to handle copyright assignment/CLA/DCA stuff
- [x] Reference it in README
- [ ] Propagate references to this document from all the other repositories
- [ ] Maybe add more details in the style guide based on past discussions, if we can find the details somewhere

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,0 +1,247 @@
+# Contributing to Nerves
+
+We're very excited that you want to contribute to the Nerves project! There are
+a lot of ways you can help out, some of which you may not have even considered
+as being a contribution to the project.
+
+Most importantly, we want the Nerves community to be welcoming and helpful to
+everyone. We expect that all members of our community follow our [Code of
+Conduct](CODE_OF_CONDUCT.md), regardless of whether the contribution is a
+discussion in a Pull Request on GitHub, a conversation in the #nerves channel on
+the [Elixir Slack], or any of the other items outlined below.
+
+[Elixir Slack]: https://elixir-slackin.herokuapp.com/
+
+## How you can help
+
+We created this document to help you quickly turn your desire to contribute into
+action. The sections below are roughly ordered by the amount of time investment
+required, but we truly value contributions of all sizes. Many small
+contributions by a large number of people will grow a stronger community and be
+more inclusive for everyone.
+
+## Try using Nerves and give us feedback
+
+Whether you're just getting started or you've been using Nerves at work for a
+long time, we would love to know where it's not meeting your needs.
+
+* Is something missing, confusing, or wrong on [our website] or our [Getting
+  Started Guide]?
+* Were you trying to create a custom [Nerves system] and it didn't work like you
+  expected?
+* Did you have fun using Nerves and you just want to tell someone about it?
+
+There are so many ways to communicate, each with advantages and disadvantages.
+We prefer communication in the following order, but we understand that you may
+have limited time to contribute and we would rather have one of the
+less-preferred items than nothing at all.
+
+One important exception is that **if you think you've found a security
+vulnerability, do not open a GitHub Issue!** Please disclose these by emailing
+[security@nerves-project.org] instead, so that we can work on releasing a fix
+before the vulnerability is disclosed publicly.
+
+1.  Pull Request on GitHub
+
+    If you're able to send us an improvement directly to the [documentation
+    source] or [website source] code, that is extremely helpful! Don't worry
+    about assigning labels, unless you also want to contribute to the [Triage
+    process] outlined below.
+
+    Also, be sure to check out the guidelines below about contributing code and
+    documentation changes.
+
+2.  Issues on GitHub
+
+    If you think you might have found a bug or documentation problem that you're
+    not sure how to solve, open an Issue on the relevant GitHub repository. The
+    Nerves project has a lot of [repositories], so if you're not sure which one
+    is most relevant, it's OK to just open it in the main [nerves repository].
+
+    If you have a question about how to use Nerves, we prefer that you use the
+    [Nerves section] on [Elixir Forum] or the #nerves channel on the [Elixir
+    Slack].
+
+3.  Questions on [Elixir Forum] (ideally in the [Nerves section])
+
+    If questions and answers are captured in a forum like this, it makes them
+    much easier for people to find later via search engine, as opposed to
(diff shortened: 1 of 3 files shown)


docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,256 @@
...
+* Did you have fun using Nerves and you just want to tell someone about it?
+
+There are so many ways to communicate, each with advantages and disadvantages.
+We prefer communication in the following order, but we understand that you may
+have limited time to contribute and we would rather have one of the
+less-preferred items than nothing at all.
+
+One important exception is that **if you think you've found a security
+vulnerability, do not open a GitHub Issue!** Please disclose these by emailing
+[security@nerves-project.org] instead, so that we can work on releasing a fix
```
fhunleth: Creating the account...

GregMefford: Thanks!

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,256 @@
...
+* Politely reminding the person that the answer will be more easily found later
+by others if it's posted to the [Elixir Forum] instead
+* Helping to get it posted to the Forum or converted to a GitHub Issue to
+capture or summarize the discussion in Slack about it
+
+## Become a Backer or Sponsor through OpenCollective
+
+The Nerves project has set up an [OpenCollective site] that allows individuals
+and companies to make one-time or recurring financial contributions to cover the
+cost of maintaining the project (e.g. paying for hosted services like AWS and
```
fhunleth: Could you delete the (e.g. ...) part? Not that it's incorrect or anything, but I hoping to keep the place for what funds were used on in one place on the OpenCollective site, so that it's easy to update.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,256 @@
...
+
+[How to Contribute to an Open Source Project on GitHub]: https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github
+
+### Copyright assignment
+
+**TODO:
+  I mostly made this up based on http://requirejs.org/docs/contributing.html.
+  Review CLA/DCO processes for other OSS projects**
+
+Licensing is very important to open-source software projects, especially for
```
fhunleth: My suggestion is that we push off the click-through agreement piece a little longer so that it doesn't hold up the other good things in this PR. How about changing the text to this for now:

Copyright and licensing are very important to open-source software projects. Nerves components fall under several licenses so please review the license for the project to which you are contributing. By contributing, you certify that you have the right to submit the changes under the project's open source license and assign copyright to the Nerves Project maintainers.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,256 @@
...
+companies who want to use that software in a commercial product. By contributing
+to the Nerves source code or documentation, you are agreeing to release your own
+intellectual property (and not the intellectual property of others) to the
+Nerves project.
+
+### Code review process
```
fhunleth: I think that we should delete this section. We probably should move it elsewhere for us, but there's a lot of the use of the term "should", so I'm not sure that it's that helpful to a contributor.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,256 @@
...
+* Each repository should have an owner and one or more maintainers.
+* At least one approval is required to merge a Pull Request.
+* Anyone on the core team can approve a Pull Request on any repository.
+
+  > NOTE: Please approve responsibly. If you're not deeply familiar with the
+  code, feel free to comment or request changes, but don't approve.
+
+* The owner of a repository should be the one to perform the merge.
+
+  > NOTE: We prefer the GitHub "Squash and Merge" button, but any method is OK.
```
fhunleth: I actually have a strong opinion against "Squash and Merge". I prefer "Rebase and Merge" so that the commit history remains linear and each commit is one logical change. Anyway, doesn't matter, since it would be good to move this section anyway.

docs/CONTRIBUTING.md
GregMefford: I think, at a minimum, we should leave the paragraph about our goal to respond to Issues and Pull Requests within a few days, so that people would know what to expect.

My intention was just to capture the rest of the details _somewhere_ because it comes up occasionally and I see it as beneficial that the core team is transparent and explicit about its workflow, so that everyone knows what to expect. If you think it's adding unhelpful clutter to this file, I don't mind putting it somewhere else. Maybe `CODE_REVIEW.md`?

docs/CONTRIBUTING.md
GregMefford: This is an example of why we should have these guidelines posted somewhere. ;)

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,230 @@
...
+prompt before the commands to be entered to make them easier to copy and paste
+into a terminal. For example:
+
+```bash
+export MIX_TARGET=rpi3
+mix firmware
+mix firmware.burn
+```
+
+When showing what we expect a person to see or do on the command-line, we will
```
fhunleth: If I understand correctly, the logic is "if output is shown, then show prompt, else don't show prompt". Therefore, if we showed one line of output in the example above, we'd require a `$` in front of all of the commands. Does this seem weird or is there more to the logic?

GregMefford: I'm not sure which "example above" you're talking about since the three lines above are all input without any output shown. My feeling is that we'd include all of the prompts or none of them in a given block of pre-formatted text.

I agree that by the logic we currently have here, you'd include all the prompts if there's only one line of output. If you thought that looked weird, you could potentially just not include the line of output if it's not necessary for understanding. I think in general it's best not to pick and choose which commands are showing output vs. only showing the commands to be run within a given block of text because someone might think there's a problem if there's a bunch of text generated for them as opposed to nothing as shown in the example. You could also split the text into multiple blocks, like "run these three commands", then "run this command, which will have output that looks like this," then "run a few more commands", etc.

If you see a way that we can make things more clear or want to change the rules, I'm totally fine with that. I just wanted to get whatever we have nailed-down so that we can build on it.

PR #211, opened 2017-10-19
Update code snippet for system dependency syntax

joelbyler:
For folks like me who like to copy and paste code snippets!  :D


mobileoverlord:
good catch 👍 
❤️

PR #213, opened 2017-10-20
Update official target list

fhunleth:
This removes Linkit, Galileo, AG150, and Alix. It adds the generic
x86_64 target.


kumekay:
Linkit 7688 is now deprecated?

fhunleth:
Right. We haven't anyone who's really been interested in maintaining it for a while. It's not like you can't use it, but it is a very time consuming target to maintain for us.

fhunleth:
We moved it here: https://github.com/mobileoverlord/nerves_system_linkit

PR #216, opened 2017-10-21
raise when artifact build encounters an error

mobileoverlord:
When a local build would encounter an error while trying to produce an artifact the command would return with a normal exit status and appear like nothing went wrong. This makes changes to not only raise but to show some of the tail of the build log.

### DIFF
--- a/lib/nerves/utils/stream.ex
+++ b/lib/nerves/utils/stream.ex
@@ -11,21 +11,39 @@ defmodule Nerves.Utils.Stream do
     GenServer.stop(pid)
   end
 
+  def history(pid) do
+    GenServer.call(pid, :history)
+  end
+
   def init(opts) do
     file = opts[:file]
+    history_lines = opts[:history_lines] || 100
+
     if file != nil do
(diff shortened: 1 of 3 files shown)


lib/nerves/utils/stream.ex
```diff
@@ -62,6 +79,17 @@ defmodule Nerves.Utils.Stream do
...
+    history = 
+      if length(s.history) >= s.history_lines do
+        [line | history] = s.history
+        history
+      else
+        s.history
+      end
+    %{s | history: history ++ [line]}
+  end
+
```
mobileoverlord: There is almost certainly a better way of doing this. Open to suggestive rewrites

fhunleth: Use Erlang's `:queue` module and keep track of the length so that you know when to remove an item for everyone that you insert?

mobileoverlord:
@fhunleth I redid this using `:queue`, ready for another peek

mobileoverlord:
We can keep a separate count. I’d rather update it now then later

PR #219, opened 2017-11-14
Unify checks for NERVES_SYSTEM & NERVES_TOOLCHAIN

dkarter:
This is the first step in drying up this code so that we can add a more
helpful message when the user simply forgot to set the `MIX_TARGET`
before running mix.firmware and related Mix tasks.

Background:

screenshot 2017-11-13 20 45 16

Thanks!


mobileoverlord:
Looks good, thanks for cleaning this up :) 👍

PR #225, opened 2017-11-21
Don't override the output_dir in the Distillery Plugin.

GregMefford:
@mobileoverlord and I are still not sure why this was added originally, but it seems to not be needed and it causes problems when you run `mix firmware` without first running `mix compile`.


mobileoverlord:
¯\_(ツ)_/¯

Issue #130, opened 2017-06-08
NERVES_SYSTEM isn't documented


fhunleth:
I noticed that neither `NERVES_SYSTEM` nor `NERVES_TOOLCHAIN` were documented or maybe I couldn't find it? This is the main way that I use Nerves, so I was wondering if these should be documented.

electricshaman:
I would vote no until we fix the weird discrepancies between NERVES_SYSTEM and the "new way."  You had previously asked me to look into a wrong set/unset message but what I uncovered was a litte more serious than just fixing the message itself

GregMefford:
@electricshaman I'm curious if you think the situation has changed at this point, with the updated docs here: https://hexdocs.pm/nerves/systems.html#customizing-your-own-nerves-system

I still want to spend some time with @fhunleth as well, understanding in more detail how his preferred workflow differs, so we can fix this documentation bug.

electricshaman:
I'll take a look this weekend!

electricshaman:
@GregMefford Looks good!  👍

ConnorRigby:
It seems we have docs for this now, but they aren't included in hexdocs. Is there a reason for this?

PR #230, opened 2017-12-12
Document nerves.precompile Call Tree

michaelkschmidt:
There are quite a few moving pieces in nerves.precompile, and this is a rough pass at documenting them.

Let me know if this should be captured in a different place.

### DIFF
--- a/docs/Internals.md
+++ b/docs/Internals.md
@@ -51,9 +51,16 @@ to be a "10,000 ft" overview.
 
 # Key Files/Variables
 
+The following are the key parts of the bootstrap.  Note that NERVES_SYSTEM and 
+NERVES_TOOLCHAIN can be defined before running `mix firmware` to point to a 
+trusted decompressed system or toolchain. This is useful in situations where 
+you produce a system directly using Buildroot and want to force Nerves to use it.
+
 NERVES_SYSTEM 
   * Path to the nerves_system_* folder
   * Has to be defined at Nerves.Env.bootstrap() or system blows up
+  * Exists only if the system dependency is being included from a source other than hex.
+    * When a system is being sourced from hex, it will attempt to place the uncompressed artifact in the global path located at ~/.nerves/artifacts or $NERVES_ARTIFACTS_DIR
 
 NERVES_TOOLCHAIN 
   * Path to the toolchain
--- a/docs/Internals.md
+++ b/docs/Internals.md
@@ -1,0 +1,69 @@
+# Internals
+
+The nerves bootstrapping process has several steps.  Its goal is to locate 
+the "system", compile it, and use the compiled system to setup the cross 
+compile environment.
+
+## Call Tree
+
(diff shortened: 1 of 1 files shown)


docs/Internals.md
```diff
@@ -0,0 +1,69 @@
...
+  * Has to defined at Nerves.Env.bootstrap() or systems blows up
+
+nerves_env.exs
+  * Sets the cross compile flags
+  * NERVES_SYSTEM and NERVES_TOOLCHAIN must be defined prior
+  * Everything thing run after will try to cross compile
+
+nerves_system/.nerves/artifacts/nerves_system_*
+   * "package" directory
+   * Gets fetched from nerves_env.exs artifact_url
```
mobileoverlord: This directory is local as stated above only if the system dependency is being included from a source other than hex.
When a system is being sourced from hex, it will attempt to place the uncompressed artifact in the global path located at `~/.nerves/artifacts` or  `$NERVES_ARTIFACTS_DIR`

docs/Internals.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Key Files/Variables
+
+NERVES_SYSTEM 
+  * Path to the nerves_system_* folder
+  * Has to be defined at Nerves.Env.bootstrap() or system blows up
+
+NERVES_TOOLCHAIN 
+  * Path to the toolchain
+  * Has to defined at Nerves.Env.bootstrap() or systems blows up
```
mobileoverlord: Maybe this is a good place to discuss that these could be set in the overall env for any situations where you want ot point to a trusted decompressed system or toolchain that you trust. This is useful in situations where you produce a system directly using Buildroot and want to force Nerves to use it.
Would solve https://github.com/nerves-project/nerves/issues/130

mobileoverlord:
I think that this is a great doc indeed. Thanks for getting it started @michaelkschmidt . I know that Mix and integrating with in it can be confusing at first. Thanks! I added two comments which I feel could really land it.

GregMefford:
I love that you took the time to write this up! ❤️ 
Several times, I've scratched out notes like this while trying to figure out how things work and how to make a change.

I have some minor nits to pick when I have time to go through this in more detail, but don't let that hold up getting this initial thing merged - we can always iterate and improve. Maybe exposing more people to how it works will also spark some improvements in how it works, because it seems to have grown over time to be more complicated than it needs to be, IMO.

michaelkschmidt:
@mobileoverlord thanks for the feedback.  Please let me know if you thought of anything else

On a side note, it seems like there is a missing piece to this setup:  I think we need to add a hook that will run a mix task before the cross compile env is setup.  The immediate use is to do `npm install` of any web related items. This has to be run before the cross compile is setup because any "native" libraries must still run on the host

mobileoverlord:
I’ll push tho out to hex docs with some upcoming changes. We should discuss further how to execute mix tasks before Nerves.env.

PR #231, opened 2017-12-14
Clarify linux and busybox configuration in Nerves System Shell

GregMefford:
We've had a few people (myself included) recently confused by this documentation.
I've tried to make the documentation as good as it can be, given how the process actually works.

@fhunleth I'm not much of a Makefile wizard, but do you see an easy way to make these steps "just work" more transparently, so people don't need to manually save a defconfig and copy the result to the right place? It seems like with all of the official systems, the paths will be consistent, so it would only be "unsafe" if they're doing something unusual in a custom system. Even in that case, they'd likely have source control in place that could allow them to put the file where they actually intended and un-do the damage to any accidentally trampled files.

### DIFF
--- a/docs/Systems.md
+++ b/docs/Systems.md
@@ -437,25 +437,50 @@ appropriate steps below:
 
 2. After `make linux-menuconfig`:
 
+    Assuming you're using the Nerves System Shell via Docker on a non-Linux host
+    and your custom system source directory is called `custom_rpi3`, you'll need
+    to do something like the following (the version identifiers might be
+    different for you).
+
     ```bash
     make linux-savedefconfig
-    cp build/linux-x.y.z/defconfig <your system>/linux-x.y_defconfig
+    cp build/linux-04c8e47067d4873c584395e5cb260b4f170a99ea/ /nerves/env/custom_rpi3/linux-4.4.defconfig
     ```
 
-    If your system doesn't contain a custom Linux configuration yet, you'll need
-    to update the Buildroot configuration (using `make menuconfig`) to point to
-    the new Linux defconfig in your system directory. The path is usually
-    something like `$(NERVES_DEFCONFIG_DIR)/linux-x.y_defconfig`.
+    If you're using Linux, the destination for the `cp` will simply be something
+    like `$(NERVES_DEFCONFIG_DIR)/linux-4.4.defconfig`.
+
+    > NOTE: If your system doesn't contain a custom Linux configuration yet,
+    you'll need to update the Buildroot configuration (using `make menuconfig`)
+    to point to the new Linux defconfig in your system directory. The path is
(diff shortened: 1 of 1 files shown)


GregMefford:
Oh, and another thought is that we can/should have the Docker container just set the `NERVES_DEFCONFIG` environment variable to `/nerves/env/<system_directory>` so that we can consistently tell people to use that variable instead of reaching into `/nerves/env` in Docker and not when in native Linux. What do you think about that? I recall that post-ElixirConf, we had talked about the fact that these directories don't map to exactly where you'd expect.

fhunleth:
Regarding the -savedefconfig feature, try `make linux-update-defconfig` and `make busybox-update-config`. See:

http://nightly.buildroot.org/manual.html#customize-store-package-config

fhunleth:
Regarding the environment variable to get to the system directory, I'm not sure. It seems reasonable, but I haven't thought through the ramifications. Given recent challenges with Arch and NixOS, it may be worthwhile revisiting using Docker on Linux for the system shell. That would make it easier to be consistent. 

Fwiw, I'd call the variable `NERVES_DEFCONFIG_DIR` to be consistent with the one that we already have.

GregMefford:
> Given recent challenges with Arch and NixOS, it may be worthwhile revisiting using Docker on Linux for the system shell

I thought the day would never come! 😀 
Seriously, though, I think it might be easier for everyone if the "normal" way was just to use a container whether you're on Linux or not, because it'll be easier to make it just work across distros, weird configurations, etc. I don't expect that there would be much of a performance hit at all when running Docker on Linux, as long as Docker is properly configured to not throttle your system resources unnecessarily.

> FWIW, I'd call the variable `NERVES_DEFCONFIG_DIR` to be consistent with the one that we already have.

Right, that's what I meant - I must've messed up the copy/paste.

fhunleth:
I think that I've come to the conclusion that when I'm doing Nerves system maintenance and low level Buildroot stuff, I'm always going to be using `nerves-env.sh`. For everything else, I'm using the system shell, and Docker is fine. There's enough glue in the system shell and Nerves `mix` integration that I don't think avoiding Docker on Linux aides understanding on how things work - certainly not for me.

electricshaman:
Let's do this.  Hell must have froze over.

ConnorRigby:
So I use Arch and it works just fine for me :)

Anyway Docker scares me a little bit but I guess I'll live

electricshaman:
@ConnorRigby I hear you, but I figure Docker is just emulating Linux so everything will still work for us if we want to go the manual way.  Our job will be to keep everyone honest such that everything can still be run outside of Docker without much hassle.

fhunleth:
@GregMefford Could you make a pass at this to fix the conflicts?

GregMefford:
Rebased and resolved conflicts (which were just the `$` prompts. 🙄)

PR #234, opened 2018-01-09
allow nerves_package config to override default provider

mobileoverlord:
Allows the nerves package config to define the default provider. This is useful in situations where you would want to force the package to be built in docker on linux


fhunleth:
I went back to the docs to see if we had documented providers at all. The only place that I found them is in `Internals.md`. I had been on the fence about documenting providers in the user docs, but I think this changes forces the issue by putting it in the nerves package config. Could you add some docs for it?

docs/Systems.md
```diff
@@ -177,7 +177,18 @@ The following keys are supported:
...
+    If this key is not defined, Nerves will choose a default provider
+    that should be used to build the artifact based on information about the host
+    computer that you are building on. For example, Mac OS will use
+    `Nerves.Package.Artifact.Providers.Docker` where as Linux will use
+    `Nerves.Package.Artifact.Providers.Local`. Specifying a provider module in 
+    the package config could be used to force the provider.
+
+6.  `provider_opts`: Optional - A keyword list of options to pass to the provider module.
+
+6.  `checksum`: The list of files for which checksums are calculated and stored
```
fhunleth: Don't worry about this now, but we should run a markdown linter over this file to fix things like the double "6." here. Since it will render correctly (or at least I think that it should), I'm not too worried about it now.

Issue #203, opened 2017-09-23
Retry partial downloads


fhunleth:
A few users have reported some confusion with the toolchain tarball fails to expand due to "Unexpected EOFs in archive". This is likely due to an interrupted download. At least, that's when I've seen it. Here's a snippet from @cjfreeze:
```
Chris@Hexa:~/.nerves/artifacts/nerves_system_rpi3-0.16.1.arm_unknown_linux_gnueabihf$ make
umask 0022 && make -C /mnt/c/Users/Chris/Projects/Elixir/local_pi/local_pi_fw/deps/rpi3/nerves_system_br/buildroot-2017.05 O=/home/Chris/.nerves/artifacts/nerves_system_rpi3-0.16.1.arm_unknown_linux_gnueabihf/.
>>> toolchain-external-custom  Extracting
xzcat /mnt/c/Users/Chris/Projects/Elixir/local_pi/local_pi_fw/deps/rpi3/nerves_system_br/buildroot-2017.05/dl/nerves_toolchain_arm_unknown_linux_gnueabihf-0.11.0.linux-x86_64.tar.xz | tar --strip-components=1 -C /home/Chris/.nerves/artifacts/nerves_system_rpi3-0.16.1.arm_unknown_linux_gnueabihf/build/toolchain-external-custom --exclude='usr/lib/locale/*'   -xf -
xzcat: /mnt/c/Users/Chris/Projects/Elixir/local_pi/local_pi_fw/deps/rpi3/nerves_system_br/buildroot-2017.05/dl/nerves_toolchain_arm_unknown_linux_gnueabihf-0.11.0.linux-x86_64.tar.xz: Unexpected end of input
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
make[1]: *** [/home/Chris/.nerves/artifacts/nerves_system_rpi3-0.16.1.arm_unknown_linux_gnueabihf/build/toolchain-external-custom/.stamp_extracted] Error 2
make: *** [_all] Error 2
```
One possible implementation would be to verify SHA256's of the downloaded tarballs in a similar manner to Buildroot.

mobileoverlord:
I’ve been working through this issue and was curious specifically about the title. It states that we should retry downloading the artifact if the download fails. I could see how in the current workflow that this might be desirable considering that in most circumstances Nerves env will attempt to build the artifact if there is an issue with the download.

[...]

Issue #202, opened 2017-09-22
Automatically check for Nerves archive updates


fhunleth:
It would be really nice if the Nerves archive could automatically check for an update in a similar way as the hex archive does it.

GregMefford:
I think if we do #204, this will will be much easier because the `nerves_bootstrap` archive would be a separately-versioned Hex package that we could potentially install at a particular version instead of only the latest release.

mobileoverlord:
Now that `nerves_bootstrap` is on hex we should be able to make a call to resolve the latest version from the repository periodically and present a message that an update is available.

mobileoverlord:
Fixed in v0.9.0

Issue #182, opened 2017-08-16
Add a Mix task for packaging a Nerves system artifact


GregMefford:
Currently, it's kind of complicated to create a `.tar.gz` file for a custom system that can be uploaded to an HTTP cache location.

I propose that we add a `mix system.artifact` or `mix nerves.system.artifact` task that would build and package the artifact, then tell you the output location where it was placed, so you can upload it.

I think this would live in the `Docker` and `Local` providers (not the `bootstrap` archive), so the `nerves.` prefix on the task shouldn't be necessary unless we want it to be consistent with `mix nerves.system.shell`.

mobileoverlord:
Fixed in v0.9.0
See: 
`mix nerves.artifact`

Issue #141, opened 2017-06-30
Don't build Nerves systems when artifact downloads fail


fhunleth:
In training meetings, it's not uncommon for someone to have the artifact download fail when building for the first time. It's probably due to us overloading the network. This triggers a Docker build of the system. Since they've never used Nerves before, they don't know that this isn't part of the normal process. When it works, it takes a long time, but usually something isn't installed for the Docker build and the error messages get to be pretty confusing.

I'd really like it if we could download failures error out and let the user fix their network, wait, etc, and try again.

What does everyone think of marking official system releases as "never build"? Obviously, it would be good to have a way to turn this off, but it seems like would be a separate step.

Or, how about making system building "opt-in"? I.e., systems never get locally built until the user enables it in their config.exs. I sort of like this since it seems like we could make the artifact download failure error message say something really helpful and instruct them that building locally is an option. When it does this, the "opt-in" part is the user acknowledging that a really big operation that involves Docket, etc. is ok with them. That way there won't be any surprises.

GregMefford:
New users may still follow the instructions in the error message and do the local build since they don't know any better, but I think I like that path better than having it just build it without asking.

mobileoverlord:
I think that a better choice would be to isolate the commands that perform network operations from compile operations. 

We should alias `deps.get` so we can perform HTTP provider operations following dependency resolution and output information regarding the download status of artifacts. If there is a problem downloading, you can just continue to call `mix deps.get`. If an artifact for a Nerves package was not downloaded during `deps.get`, it would be built during a call to `mix compile`.

If the operations are properly organized, I don't see a reason to prompt or offer an "opt-in". A system output directory needs to be present at the end of the `deps.precompile` phase otherwise mix needs to be halted to prevent building dependencies incorrectly.

GregMefford:
That seems like a good way to go, since downloading the artifact is logically a part of the deps-getting process. There's still a chance that the download fails during `dep.get`, in which case we should have a prominent warning or maybe even an error, saying that an artifact URL was given but it failed to download. If they know what they're doing and intend to build locally, there shouldn't be an artifact URL present at all, right?

mobileoverlord:
Right, and if there isn't one present at all `deps.get` can warn them that those artifacts are going to be built when calling `compile`

mobileoverlord:
Fixed in v0.9.0
Starting in v0.9.0 artifacts are downloaded after `mix deps.get`

PR #235, opened 2018-01-09
Refactor providers

mobileoverlord:
This is a big PR so I'll break down the gist of it. Here is what it does

Mix Tasks:
* Add `nerves.artifact.archive` to create an artifact archive and artifact checksum.
* Add `nerves.artifact.get` to call the resolver and attempt to pull the artifact from a remote location.

Functionality:
* `HTTP` is no longer a provider and has been renamed to `Nerves.Package.Artifact.Resolver`
* Providers are no longer a list they are now a single tuple of `{mod, opts}`. Packages will only have a single provider. This used to be a list and we would always try `HTTP` first before using a build provider. This was always kind of weird and never really felt right. This pulls the network operations apart from the build operations. 
* `nerves.artifact.get` will validate the archive checksum against the value of `artifact_checksum` key in `nerves_package` config. 

Organization:
* Changes the namespace for `Nerves.Package.Providers` to `Nerves.Package.Artifact.Providers`
* Moves artifact related functions out of `Nerves.Package` and into `Nerves.Package.Artifact`

When I started this task I tried to make small steps but it seemed that I couldn't pull this apart in any reasonable fashion and essentially I am one step away from squashing it all into a single commit with the title "Change all the things!"

Depends on 
https://github.com/nerves-project/nerves_bootstrap/pull/6

Fixes:

https://github.com/nerves-project/nerves/issues/203
https://github.com/nerves-project/nerves/issues/202
https://github.com/nerves-project/nerves/issues/182
https://github.com/nerves-project/nerves/issues/141


mobileoverlord:
One question that came up was how to handle existing projects that have not added the alias `deps.get`. In the current standing, it will build the artifact, but, I wanted to figure out how to inform the user that they should add the alias.

In researching this I have found that we can pull the aliases from `Mix.Project.config()[:aliases]` and inspect hem to see if it includes one for `deps.get` and ensure that it contains `nerves.deps.get`

I think we should make this check at `nerves.env` start, which happens before any code is compiled

Issue #143, opened 2017-07-03
Artifact download fails from https://raw.githubusercontent.com


GregMefford:
This issue was reported by um404 in the Slack channel.

It appears that, in this case, the problem is a missing `CONTENT-DISPOSITION` header, but this error message also isn't helpful. We should improve the error-handling as well as solving this issue with a missing header.

### Current behavior

```bash
$ MIX_TARGET=zynq mix firmware

Mix environment
  MIX_TARGET:   zynq
  MIX_ENV:      dev

warning: variable "target" is unused
  mix.exs:59

==> nerves_system_zynq
Generated nerves_system_zynq app
|Elixir.Nerves.Package.Providers.HTTP| Downloading Artifact nerves_system_zynq-0.1.0.arm_unknown_linux_gnueabihf.tar.gz
  From Location:
    https://github.com/uldza/nerves_system_zynq/raw/master/nerves_system_zynq.tar.gz

** (EXIT from #PID<0.70.0>) an exception was raised:
    ** (MatchError) no match of right hand side value: nil
        (nerves) lib/nerves/utils/http_client.ex:59: Nerves.Utils.HTTPClient.handle_info/2
        (stdlib) gen_server.erl:601: :gen_server.try_dispatch/4
        (stdlib) gen_server.erl:667: :gen_server.handle_msg/5
        (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3

10:48:18.598 [error] GenServer #PID<0.186.0> terminating
** (MatchError) no match of right hand side value: nil
    (nerves) lib/nerves/utils/http_client.ex:59: Nerves.Utils.HTTPClient.handle_info/2
    (stdlib) gen_server.erl:601: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:667: :gen_server.handle_msg/5
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
[...]

PR #238, opened 2018-01-20
Nerves.Util.HTTPClient: Don't blow up on missing Content-Disposition.

prook:
If the Content-Disposition header is missing when downloading an artifact, infer the filename from the URL instead of blowing up. This adresses #143.

I'm not sure if using Path.basename/1 with an URL is too neanderthal, but it works.


mobileoverlord:
Thanks for taking a pass through this. Its been open for some time

mobileoverlord:
:heart: Thanks!

prook:
Proud to scratch my own itch. :)

PR #241, opened 2018-02-03
Host tools fork 

elcritch:
Here's a sketch of a proposed platform for adding extra toolchains. Instead of each "extra" toolchain defining their own platform, we have one fairly "generic" platform which has various hooks that can be configured in that extra-toolchain's `mix.exs` file. It could maybe be setup to be a behavior too. 

I tried to making a separate platform for the pru cgt toolchain, but kept running into issues where the platform module hadn't been loaded by the time that the bootstrap functions run. Creating a generic platform avoids that problem but move configuation into the `nerves_project` section in `mix.exs`. There are two example of how this could be used:

The PRU CGT Toolchain which show an example of the toolchain_extra's "api": 
- https://github.com/elcritch/toolchain_extras_pru_cgt/blob/00214767c1c95fdaed364fa890cd179cc6112969/mix.exs#L26

And a "hack" to allow the separate `pru_support` lib to modify the system env with it's dependency path. Ideally, these two libraries could be merge once the artifact.build functionality is working in pru toolchain which would be much clean. 

Still it showcases that it'd be possible to pass lambda's from the nerves_config section to the extra's toolchain platform. This allows a way to add custom toolchain behaviors. I'd prefer a proper @behavior system but not sure how to get around the bootstrap issue mentioned above. 

- https://github.com/elcritch/nerves_pru_support/blob/a5651e2abbd4b5edcf8fa083669b62a9f3c712f2/mix.exs#L49

P.S. I don't expect this to be in a mergeable state but figured a PR would be a good way to discuss the solution method.


elcritch:
@mobileoverlord

PR #243, opened 2018-02-09
Host Tools and Integration Tests

mobileoverlord:
The bulk of this PR is in tests and docs. 

Other then that the summary of changes is
* Allow `nerves.artifact` to be called for any package that defines a `Nerves.Artifact.Provider` module.
* Fixes a bug with deriving the env var used to override the artifact location. `:name` is not a key on `%Nerves.Package{}`

For example, We use `NERVES_TOOLCHAIN` and `NERVES_SYSTEM` as as special tags, but, a package with the app name `host_tool` would be able to defined by setting `HOST_TOOL` to a path.

* `Artifact.host_tuple` should only ever be `portable` for system types
* Fixes a bug with putting artifacts in the global cache where it would fail creating the directory if the directory already exists. 
* Call `platform.bootstrap` on all packages _after_ systems and toolchains. Previous behaviour was that `platform.bootstrap` was _only_ being called on the system and toolchain.
* Add `Nerves.Utils.File.tar/2` helper


mix.exs
```diff
@@ -28,7 +28,7 @@ defmodule Nerves.Mixfile do
     [
       {:distillery, "~> 1.4"},
       {:ex_doc, "~> 0.16", only: [:test, :dev]},
-      {:nerves_bootstrap, github: "nerves-project/nerves_bootstrap", only: [:test, :dev]}
+      {:nerves_bootstrap, github: "nerves-project/nerves_bootstrap", branch: "host_tools", only: [:test, :dev]}
```
fhunleth: This affects nerves_bootstrap?

lib/nerves/artifact.ex
```diff
@@ -33,11 +33,20 @@ defmodule Nerves.Artifact do
...
+    """
+  end
+  def archive(pkg, toolchain, opts) do
     Mix.shell.info("Creating Artifact Archive")
     opts = default_archive_opts(pkg, opts)
+  
     case pkg.provider do
       {provider, _opts} ->
+
+        Code.ensure_compiled(pkg.platform) |> IO.inspect
```
fhunleth: Did you want the IO.inspect on this line?

lib/nerves/artifact.ex
```diff
@@ -33,11 +33,20 @@ defmodule Nerves.Artifact do
   @doc """
   Produces an artifact which can be fetched when calling `nerves.artifact.get`.
```
fhunleth: The text and functions here are using the words "artifact" and "archive" interchangably. I believe that you wanted line 38 to read "mix archive" instead of "mix artifact", but you should probably review the lines around this comment to make sure that they're like you want.

lib/mix/tasks/nerves.artifact.ex
```diff
@@ -29,6 +29,8 @@ defmodule Mix.Tasks.Nerves.Artifact do
   @switches [path: :string]
 
   def run(argv) do
+    Mix.Task.run("nerves.precompile", [])
```
fhunleth: Add a comment here for why you're doing this?

mix.exs
mobileoverlord: See https://github.com/nerves-project/nerves_bootstrap/pull/14 for the bootstrap updates.

mobileoverlord:
I cleaned up and added a few notes where lacking. Got the tests to pass on CI.

PR #247, opened 2018-02-14
add dep opts and compilers to package struct

mobileoverlord:
Adds metadata from the dependency deceleration and compilers to the package struct to use during precompile. This also adds a test to ensure that we raise during precompile if if the project fails compile check.

Requires this PR
https://github.com/nerves-project/nerves_bootstrap/pull/21


mobileoverlord:
I updated this to only pull out options for nerves under the dep like this:

```elixir
{:my_dependency, "~> 1.0", nerves: [compile: true]}
```

Everything under `:nerves` will get pulled into the `%Nerves.Package{dep_opts: []}`

Issue #250, opened 2018-02-17
File name with full hash longer 143 characters aren't supported by ecryptfs


webengineer-max:
Filenames like .nerves_toolchain_arm_unknown_linux_gnueabihf-linux_x86_64-0.13.0-D25664C99BC7E3023A2C093C726649E7D50333B9F3887738634F124C9B5BA173.tar.xz.XXXXXX longer than 143 characters aren't supported by ecryptfs (https://bugs.launchpad.net/ecryptfs/+bug/344878). As a result I got error

```
mktemp: failed to create directory via template '/home/profi/Code/tk/deps/rpi3/nerves_system_rpi3/.nerves/artifacts/nerves_system_rpi3-portable-0.20.0/build/.nerves_toolchain_arm_unknown_linux_gnueabihf-linux_x86_64-0.13.0-D25664C99BC7E3023A2C093C726649E7D50333B9F3887738634F124C9B5BA173.tar.xz.XXXXXX': File name too long
```

during `mix firmware` inside encrypted home folder.

### Environment
* Erlang version: 20
* Elixir version: 1.6.1
* Operating system: Ubuntu 17.10
* Filesystem: ecryptfs (libecryptfs1/artful,now 111-0ubuntu5 amd64)
* Nerves Environment Info:

|nerves_bootstrap| Environment Package List

  Pkg:      nerves_toolchain_arm_unknown_linux_gnueabihf
  Vsn:      0.13.1
  Type:     toolchain
  Provider: {Nerves.Artifact.Providers.Local, []}

  Pkg:      nerves_toolchain_ctng
  Vsn:      1.3.1
  Type:     toolchain_platform
  Provider: nil

  Pkg:      nerves_system_rpi3
  Vsn:      0.20.0
  Type:     system
  Provider: {Nerves.Artifact.Providers.Local, []}

  Pkg:      nerves_system_br
  Vsn:      0.17.0
  Type:     system_platform
  Provider: nil

|nerves_bootstrap| Loadpaths Start

|nerves_bootstrap| Environment Variable List
  target:     rpi3
  toolchain:  /home/profi/.nerves/artifacts/nerves_toolchain_arm_unknown_linux_gnueabihf-linux_x86_64-0.13.1
  system:     /home/profi/.nerves/artifacts/nerves_system_rpi3-portable-0.20.0
  app:        /home/profi/Code/tk/apps/fw

|nerves_bootstrap| Loadpaths End

### Current behavior

[...]

PR #251, opened 2018-02-19
Change artifact download to use short checksums

mobileoverlord:
Instead of using the full checksum for the artifact download file name we will use only the first 7 similar to that of the git default. Since the artifact name also contains the version number, uniqueness isn't going to be an issue.

### DIFF
--- a/lib/nerves/artifact.ex
+++ b/lib/nerves/artifact.ex
@@ -7,6 +7,7 @@ defmodule Nerves.Artifact do
   alias Nerves.Artifact.{Cache, Providers}
 
   @base_dir Path.expand("~/.nerves/artifacts")
+  @checksum_short 7
 
   @doc """
   Builds the package and produces an  See Nerves.Artifact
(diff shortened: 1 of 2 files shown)


fhunleth:
Where did you land on this one and not breaking all current systems that use the entire hash string?

mobileoverlord:
I updated the site expander to add both versions for now. It will first try the short sum and then fall through to the long one. I also added some comments to the code regarding cleaning this up sometime following 1.0 since the plan is to only use short sums moving fwd.

Fixes https://github.com/nerves-project/nerves/issues/250

```
Resolving Nerves artifacts...
  Resolving nerves_system_bbb
  => Downloading https://github.com/nerves-project/nerves_system_bbb/releases/download/v0.20.0/nerves_system_bbb-portable-0.20.0-F7599E9.tar.gz
  => Downloading https://github.com/nerves-project/nerves_system_bbb/releases/download/v0.20.0/nerves_system_bbb-portable-0.20.0-F7599E9B2F6A9BCAE5B9E2B98A3440CDFEA6B0FA347F290605C4111E11289C8F.tar.gz
|==================================================| 100% (112 / 112) MB
```

PR #252, opened 2018-02-21
Updated UI docs with shoehorn

konstantinzolotarev:


mobileoverlord:
Awesome, nice find. After we put out 1.0 we can take a pass through and consolidate this again. 👍

PR #256, opened 2018-03-06
Allow for future docker versions to work with mix firmware

mattludwigs:
Currently, docker's edge channel is running version `18.03.00`, which I have installed locally. I was trying to be run `mix firmware` to build my company's firmware and was not able to. I was getting an error message telling me that I did not have docker installed locally, but I knew that was false and verified. After searching the repo I found: https://github.com/nerves-project/nerves/pull/85. Which suggests the possibility of just setting the version to `>= 0.0.0` or `>= 17.0.0`. So I updated the version to string to include `>= 17.0.0`. I am sure if we want more constraints on the versions that are allowed for docker or not, but I wanted to get the conversation started.

While most people are not on the edge channel, I would suspect that a version 18 release will happen at some point and this code will need to change anyway. Further, others who are on the edge channel (some co-workers of mine) will probably run into this error as well. After making this update the docker version error went away and I am able to run `mix firmware` successfully.

### DIFF
--- a/lib/nerves/artifact/providers/docker.ex
+++ b/lib/nerves/artifact/providers/docker.ex
@@ -62,7 +62,7 @@ defmodule Nerves.Artifact.Providers.Docker do
   alias Nerves.Artifact.Providers.Docker
   import Docker.Utils
 
-  @version "~> 1.12 or ~> 1.12.0-rc2 or ~> 17.0"
+  @version "~> 1.12 or ~> 1.12.0-rc2 or >= 17.0.0"
   @tag "nervesproject/nerves_system_br:latest"
 
   @dockerfile File.cwd!()


mix.lock
```diff
@@ -2,5 +2,5 @@
   "distillery": {:hex, :distillery, "1.5.2", "eec18b2d37b55b0bcb670cf2bcf64228ed38ce8b046bb30a9b636a6f5a4c0080", [:mix], [], "hexpm"},
   "earmark": {:hex, :earmark, "1.2.4", "99b637c62a4d65a20a9fb674b8cffb8baa771c04605a80c911c4418c69b75439", [:mix], [], "hexpm"},
   "ex_doc": {:hex, :ex_doc, "0.18.3", "f4b0e4a2ec6f333dccf761838a4b253d75e11f714b85ae271c9ae361367897b7", [:mix], [{:earmark, "~> 1.1", [hex: :earmark, repo: "hexpm", optional: false]}], "hexpm"},
-  "nerves_bootstrap": {:hex, :nerves_bootstrap, "1.0.0-rc.0", "38fc74d585fea1973dcca1235ddeecf3a94696bfb84974134adf6fcc13312057", [:mix], [], "hexpm"},
+  "nerves_bootstrap": {:hex, :nerves_bootstrap, "1.0.0-rc.0", "b279c74035a612c40ee4acb34c21e5082bd604410e30ea9d030fec8cede4a027", [:mix], [], "hexpm"},
```
fhunleth: This is a slightly disturbing change. @mobileoverlord - Did `nerves_bootstrap` 1.0.0-rc.0 get pushed twice to hex.pm?

@mattludwigs Could you update your PR to remove the change to mix.lock just so we have a clean commit?

mobileoverlord:
👍 I can't find any changelog notes for what is different in 18, but, like you said, we can act accordingly if anything breaks instead of inhibiting it from the start.

Issue #259, opened 2018-03-07
Artifact sites should allow request headers and params to be specified


mobileoverlord:
Currently, artifact_sites helpers only pass the type and the string value. 

For example:
`{:prefix, "https://my_server.com/fw_files/"}`

This falls short when trying to pass authentication tokens or any additional information in the query params or request headers.

We should allow additional options to be passed through for adding headers and params.

Lets say we need to use token auth in the query params.
`{:prefix, "https://my_server.com/fw_files/", params: %{"auth_token" => System.get_env("AUTH_TOKEN")}}`

We can also support headers in a similar fashion.
`{:prefix, "https://my_server.com/fw_files/", headers: %{"authorization" => "type: basic 12345"}}`

mobileoverlord:
This will also work for the `:github_releases` helper since it expands to `:prefix`

PR #261, opened 2018-03-09
allow artifact_site helpers to take headers and query params

mobileoverlord:
Fixes https://github.com/nerves-project/nerves/issues/259

This PR adds:

* Artifact site helpers can pass options for the resolver.
  * Add custom request headers using `{:prefix, "http://server.com/", headers: [{"My Header", "1234"}]}`
  * Add custom query parameters using `{:prefix, "http://server.com/", query_params: %{"key" => "value"}}`

* Added test server and increased coverage on the artifact resolver tests


mattludwigs:
@mobileoverlord and @fhunleth I have been testing this PR locally this morning and I have not been having too much luck. After some research, it appears that GitHub does not all private repo release assets to be downloaded from `https://github.com` and we might have to go through the `https://api.github.com` URL.

I have tried passing headers and adding the query params using both the PR code and curl to `github.com`, but no luck yet. My token seems to work best with `api.github.com`.

The bummer is I have only found dated information (around 2015) or just little bits of information here or there, but nothing concrete to confirm what I said above is actually true, and I could just be making silly mistakes on my end.

mobileoverlord:
It appears that the problem with going through the github api is that it expects everything to be referenced by an internal id for traversal. This would mean that we would need to have a conversation with github before being able to fully resolve the final url of the asset.

mattludwigs:
Yeah, I saw that and that is not ideal. I am going to do some more testing here in a little bit to see if there a way around having to make multiple requests and to verify that I am not doing something wrong on my end.

mobileoverlord:
Updated this PR to further expand on the initial work of adding support for other options that would go into resolving the final artifact. The latest commit adds initial support for fetching artifacts using the Github API. In order to decode the responses from Github, Nerves has an optional dependency on `jason`. You will need to specify this in your project dependencies:

```elixir
{:jason, "~> 1.0"}
```

If you are already including a JSON library like Poison for example, you can use it instead by updating your `config.exs`

```elixir
config :nerves, json_codec: Poison
```

Finally will need to supply your username, a personal access token, and the tag of the release. Here is how you would define the `artifact_site`

```elixir
{:github_api, "github_org/my_custom_system", username: System.get_env("GITHUB_USER"), token: System.get_env("GITHUB_TOKEN"), tag: @version}
```

I have tested fetching through this mechanism using public repositories but not private ones.

PR #263, opened 2018-03-09
Provide a better error message on Docker failures

fhunleth:
Before this change, the error message says to look at build.log.
It's not obvious that build.log is buried under the deps directory, so
to make this easier, the change tails the build.log and gives the full
path.


mobileoverlord:
Thanks 👍

PR #264, opened 2018-03-16
add nice error messages to :github_api bases artifacts

mattludwigs:


lib/nerves/artifact/resolvers/github_api.ex
```diff
@@ -44,4 +47,48 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
       %{"url" => url} -> {:ok, url}
     end
   end
+
+  defp get_username(opts) do
+    case Keyword.fetch(opts, :username) do
+      {:ok, nil} ->
```
mobileoverlord: Connor is speaking about this not matching. Should be just `:error`

lib/nerves/artifact/resolvers/github_api.ex
```diff
@@ -44,4 +47,48 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
...
+        For correctly setting up your environment please see the documentation for the artifact you are
+        trying to download.
+        """)
+      {:ok, username} -> username
+    end
+  end
+
+  defp get_token(opts) do
+    case Keyword.fetch(opts, :token) do
+      {:ok, nil} ->
```
mobileoverlord: same here `:error`

ConnorRigby:
Sorry, kind of a premature enter on that comment, instead of `{:ok, nil`}`, it should be `:error`.
or you can try `Keyword.get(opts, key)` which will return your value or `nil`

lib/nerves/artifact/resolvers/github_api.ex
adkron: What if someone explicitly sets the option to `nil`? We might want both.

mobileoverlord:
@adkron I would say that `username` `token` and `tag` are all required fields.

adkron:
I think that switching to `:get` is nice because it handles the explicit `nil`. I agree that they are required, but someone could put in a `nil` value and we need to handle that just like there isn't a value.

Let's add some typespecs that show the fields are required. I know that @mattludwigs knows how.

mattludwigs:
@ConnorRigby @mobileoverlord I can update that, however, if someone sets:

```
{username: System.get_env("GITHUB_USER"), token: System.get_env("GITHUB_TOKEN")}
```

But does not actually set those fields in their environment, then `Keyword.fetch` will return `{:ok, nil}`, not just `:error`, so I think we should account for both cases.

`Keyword.get` might be a better choice here because either way that will just return `nil` and we can match on that instead.

mattludwigs:
@mobileoverlord I think `docs/Systems.md` needs to be updated too to reflect the `:github_api` addition. I can do that too in this branch.

lib/nerves/artifact/resolvers/github_api.ex
```diff
@@ -44,4 +46,68 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
...
+      username ->
+        username
+    end
+  end
+
+  defp get_token(opts, artifact_name) do
+    case Keyword.get(opts, :token) do
+      nil ->
+        Mix.raise("""
+        GitHub token not set for artifcat: #{artifact_name}
```
paulstatezny: Artif cat! 😸 

Sorry, had to!

mobileoverlord:
❤️ Thanks @mattludwigs

PR #269, opened 2018-04-06
Allow for multiple rootfs additions.

ConnorRigby:
This allows for a user to supply a list of additions _or_ a single string.

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -92,6 +92,9 @@ defmodule Mix.Tasks.Firmware do
     rootfs_overlay =
       case firmware_config[:rootfs_overlay] || firmware_config[:rootfs_additions] do
         nil -> []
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/firmware.ex
```diff
@@ -92,6 +92,9 @@ defmodule Mix.Tasks.Firmware do
     rootfs_overlay =
       case firmware_config[:rootfs_overlay] || firmware_config[:rootfs_additions] do
         nil -> []
+        overlays when is_list(overlays) ->
+          Enum.map(overlays, fn(overlay) -> ["-a", Path.join(File.cwd!(), overlay)] end)
```
fhunleth: Minor, but you can use `Enum.flat_map/2` and do this in one step.

ConnorRigby: ah yes, i always forget about `flat_map/2`

PR #271, opened 2018-04-22
use find_executable instead of calling out to which / where

mobileoverlord:
During training at ElixirConf EU we have found that Node.js was interfering with calls to `System.cmd "which", ["some_exec"]`. It seems that it shims `which`.

https://github.com/asdf-vm/asdf/issues/284#issuecomment-366010826

Sigh.
This changes the script to use find_executable instead.


fhunleth:
I'm not going to forget this issue in a while... I assume that we'd still hit the Node issue on a later `System.cmd` for people who are affected by it?

PR #272, opened 2018-04-22
Fix regression with checking downloads directory for artifacts

mobileoverlord:
The behavior for downloading artifacts neglected to consult the download directory  before trying to download from artifact sites. This adds a test and fixed the regression.

### DIFF
--- a/test/nerves/cache_test.exs
+++ b/test/nerves/cache_test.exs
@@ -60,4 +60,24 @@ defmodule Nerves.CacheTest do
       assert_received({:mix_shell, :info, [^output]})
     end)
   end
+
+  test "corrupted downloads are validated on get and removed" do
+    in_fixture("system", fn ->
+      File.cwd!()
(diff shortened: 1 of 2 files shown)


test/nerves/cache_test.exs
```diff
@@ -32,4 +32,32 @@ defmodule Nerves.CacheTest do
       assert Cache.valid?(package)
     end)
   end
+
+  test "dl folder is queried prior to calling resolver" do
```
fhunleth: Is it easy to test the case where the file exists, but has a bad checksum? That seems useful to verify. I'm on the fence on what happens on a failure, but slightly leaning towards Buildroot's policy of deletion.

mobileoverlord:
@fhunleth I updated this to verify the file on `nerves.artifact.get`. If a file exists and it does not pass validation, the code will remove it from the dl directory.

PR #277, opened 2018-05-01
Synchronize user:group permissions between docker, host, and volumes

mobileoverlord:
Docker currently runs as the root user when building Nerves systems. This is problematic when it is required to write some files back to bind mounted host directories such as `deps/nerves_system_br`. This causes permissions issues when cleaning dependencies.

This PR configures the container to override the user id and group id to match that of the executing user.

### DIFF
--- a/lib/nerves/artifact/build_runners/docker.ex
+++ b/lib/nerves/artifact/build_runners/docker.ex
@@ -125,7 +125,13 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
 
     mounts = Enum.join(mounts(pkg), " ")
     ssh_agent = Enum.join(ssh_agent(), " ")
-    cmd = "docker run --rm -it -w #{@working_dir} #{mounts} #{ssh_agent} #{image}"
+
+    cmd =
+      "docker run --rm -it --user #{uid()}:#{gid()} -w #{@working_dir} #{mounts} #{ssh_agent} #{
+        image
+      }"
+
+    set_volume_permissions(pkg)
     Mix.Nerves.Shell.open(cmd, initial_input)
   end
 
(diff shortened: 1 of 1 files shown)


lib/nerves/artifact/build_runners/docker.ex
```diff
@@ -125,7 +125,14 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
 
     mounts = Enum.join(mounts(pkg), " ")
     ssh_agent = Enum.join(ssh_agent(), " ")
-    cmd = "docker run --rm -it -w #{@working_dir} #{mounts} #{ssh_agent} #{image}"
+
+    cmd =
+      "docker run --rm -it --user #{uid()}:#{gid()} -w #{@working_dir} #{mounts} #{ssh_agent} #{
+        image
+      }"
+      |> IO.inspect()
```
fhunleth: Do you want to leave in the IO.inspect?

ConnorRigby: FWIW i really like Credo, and it can lint/catch that

lib/nerves/artifact/build_runners/docker.ex
```diff
@@ -201,6 +212,41 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
...
+      {result, _} ->
+        Mix.raise("""
+        The Nerves Docker build_runner encountered an error while setting permissions:
+
+        #{inspect(result)}
+        """)
+    end
+  end
+
+  defp uid() do
```
fhunleth: How about `System.cmd("id", ["-u"])`?

mobileoverlord:
Updated to address the comments.

lib/nerves/artifact/build_runners/docker.ex
```diff
@@ -201,6 +211,39 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
     end
   end
 
+  defp set_volume_permissions(pkg) do
+    {_dockerfile, image} = config(pkg)
+
+    args =
```
fhunleth: I'm good with this if you add a comment here to explain what's going on with the args and `--rm` and `chown`.

PR #281, opened 2018-05-16
Fix style guide inconsistency

GregMefford:
I noticed while reviewing #280 that the paragraph says that we _will_ use prompts and the examples and actual documentation _don't_ use prompts. I hope this is the last time we switch back and forth. 😿


mobileoverlord:
Thanks, good catch

Issue #183, opened 2017-08-16
Add a Mix task to return the full path to the firmware file


GregMefford:
The goal is to eliminate the need for things like this: https://github.com/fhunleth/nerves_firmware_ssh/blob/master/upload.sh#L25

Possible solutions include:
* New flags to `mix firmware` (e.g. `mix firmware --quiet --path`) that would ensure the firmware file exists and return its full path
* New Mix task (either in `bootstrap` or `nerves`) that would simply return where the firmware _should_ be, but not actually build it (e.g. `mix nerves.info --firmware-path`)

GregMefford:
@ConnorRigby Since you closed #283, should we close this too, or should we still come up with a solution? I realize I created this issue, but I think it was based on a Slack discussion we were having where we were thinking it would be nice to have Mix do this kind of stuff for us.

ConnorRigby:
Missed this somehow. I think this is safe to be closed. Anyone feel free to reopen if this is needed.

PR #283, opened 2018-05-17
Firmware task update

ConnorRigby:
I just had a use case for this, but unfortunately, the Erlang shell information gets printed to the screen and as far as i know that can't be disabled so i don't think adding this flag will be super useful..

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -22,6 +22,7 @@ defmodule Mix.Tasks.Firmware do
   def run(args) do
     preflight()
     debug_info("Nerves Firmware Assembler")
+
     if "--image-path" in args do
       config = Mix.Project.config()
       images_path = images_path(config)
@@ -120,7 +121,6 @@ defmodule Mix.Tasks.Firmware do
   end
 
   defp standard_fwup_variables(config) do
-    config = Mix.Project.config()
     # Assuming the fwup.conf file respects these variable like the official
     # systems do, this will set the .fw metadata to what's in the mix.exs.
     [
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/firmware.ex
```diff
@@ -19,31 +19,36 @@ defmodule Mix.Tasks.Firmware do
...
+  def run(args) do
     preflight()
     debug_info("Nerves Firmware Assembler")
 
-    system_path = check_nerves_system_is_set!()
+    if "--image-path" in args do
+      config = Mix.Project.config()
+      images_path = images_path(config)
+      otp_app = config[:otp_app]
+      IO.inspect(fw_path(images_path, otp_app))
```
GregMefford: Is this `IO.inspect` here on purpose? I'm a little unclear on what the purpose of this `--image-path` flag is. The docs also don't describe that this flag is here and what it does.

ConnorRigby: I put `IO.inspect`  so that the task could be used to get the data from other Elixir code _and_ from the shell. The initial reason for this was to stop the shell script in Nerves Firmware SSH from doing this all manually. Unfortunately since Erlang spits stuff to the terminal no matter what (as far as i know), that isn't very useful. We can probably turn it to an `IO.puts` or `Mix.shell.info`. (if we decide this is actually needed that is.)

ConnorRigby: FWIW the issue brought this up is #183

ConnorRigby:
This is outdated now, and i don't think we actually need this feature anymore. Closing.

PR #286, opened 2018-05-26
fixed library name in install doc

lenhart:
Corrected library name for install on debian/ubuntu systems .

https://packages.debian.org/stretch/libwxgtk3.0-dev


mobileoverlord:
Thanks!

PR #289, opened 2018-06-01
Do not fetch artifacts on deps.get that are set in the env

mobileoverlord:
When calling `mix deps.get` do not attempt to fetch the artifacts if the artifact path is being overridden by setting its env variable such as `NERVES_SYSTEM`

### DIFF
--- a/test/nerves/cache_test.exs
+++ b/test/nerves/cache_test.exs
@@ -80,4 +80,22 @@ defmodule Nerves.CacheTest do
       refute File.exists?(dl_path)
     end)
   end
+
+  test "skip fetching packages that have paths set in the env" do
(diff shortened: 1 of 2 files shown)


ConnorRigby:
Coincidentally i as about to raise an issue about this because it sort of annoyed me. :+1:

test/nerves/cache_test.exs
```diff
@@ -80,4 +80,22 @@ defmodule Nerves.CacheTest do
       refute File.exists?(dl_path)
     end)
   end
+
+  test "ship fetching packages that have paths set in the env" do
```
ConnorRigby: ~ship~ skip?

PR #292, opened 2018-06-24
Fix documentation for saiving the Linux config.

cdegroot:
Also bumped the target file to the current  kernel version for clarity.

### DIFF
--- a/docs/Systems.md
+++ b/docs/Systems.md
@@ -443,7 +443,7 @@ appropriate steps below:
 
     ```bash
     make linux-savedefconfig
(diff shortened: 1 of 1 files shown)


ConnorRigby:
I _only_ use the old method of building and have never used the new method, so i can't confirm this works. For what it's worth the old method i only ever have to `make linux-savedefconfig`. 

@mobileoverlord  or @GregMefford should probably have a look at this.

mobileoverlord:
This is great. Thanks!

PR #294, opened 2018-06-26
mix format

tmecklem:
Missed `mix format` commit in the wsl firmware.burn changes


tmecklem:
Merging without review since this is just changes from the formatter

PR #295, opened 2018-07-06
Add build runner option for passing args to make

mobileoverlord:
This PR allows `build_runner_opts` to be used to pass additional arguments to the call to make.

For example:

The `kiosk_system_*`  Nerves packages build QtWebEngine which consume a lot of memory and cause the build to fail in an obscure way. To address the problem we need to reduce the number of `PARALLEL_JOBS` to 8.

Here is how we would pass the args through the mix.exs file
```
# mix.exs
defp nerves_package do
  [
    # ...
    build_runner_opts: [make_args: ["PARALLEL_JOBS=8"]],
  ]
end
```

### DIFF
--- a/docs/Systems.md
+++ b/docs/Systems.md
@@ -221,6 +221,22 @@ The following keys are supported:
 
 6. `build_runner_opts`: Optional - A keyword list of options to pass to the build_runner module.
 
+    `make_args:` - Extra arguments to be passed to make. 
+    
+    For example:
+    
+    You can configure the number of parallel jobs that buildroot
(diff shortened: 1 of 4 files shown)


fhunleth:
Fwiw, the example in your PR description looks super-useful. I think that it should be copy/pasted into the docs for `make_args`.

mobileoverlord:
Thanks for the feedback. I added the example to the docs for 

`lib/nerves/build_runners/local.ex`
`lib/nerves/build_runners/docker.ex`
`docs/Systems.md`

Issue #324, opened 2018-08-23
Proposal: make firmware.burn work in a Nix shell


ejpcmac:
I’m currently migrating all my development workflow to [Nix](https://nixos.org/nix/). It has the advantage of letting you define a `shell.nix` in your project that lists dependencies needed to setup your build environment.

I’ve come up to be able to build Nerves firmware in pure Nix shells—that is, with the user environment being reset before to setup the shell. The issue is now with running `mix firmware.burn`.

On Linux, `mix firmware.burn` needs `sudo`. The problem is that currently, Nix does not `setuid`, so even with `sudo` installed it cannot work. More: even if `sudo` were working, it would reset the environment, thus `fwup` would not be available—since it is made available locally by the Nix shell and not system-wide.

Currently, the only way to flash a firmware is to install `fwup` system-wide and use it with the system `sudo`.

I have no ready-to-implement solution but I will think about it. If you have some ideas too, I’m all ears. I open this issue today to track it and start a brainstorming.

## Ideas (edit)

* We may abandon the idea of `mix firmware.burn` running in pure Nix shells are they are only intended to test buildability and not making side effects.
* Doing this, the only remaining issue is the unavailability of `fwup` system-wide. Instead of doing `sudo fwup <file>`, we should do `sudo $(which fwup) <file>`. We may use `which` also to find `ssh-askpass`. It would enhance the UX on distributions like Fedora where it is not installed at `/usr/bin/ssh-askpass`.

PR #325, opened 2018-08-23
Make firmware.burn work with fwup from a Nix shell

ejpcmac:
Following #324, I’ve made a tiny patch that enables to use `mix firmware.burn` in a Nix shell where `fwup` is available while not installed system-wide. It uses `which` to find its path before calling `sudo`.

### DIFF
--- a/lib/mix/tasks/firmware.burn.ex
+++ b/lib/mix/tasks/firmware.burn.ex
@@ -112,7 +112,7 @@ defmodule Mix.Tasks.Firmware.Burn do
                 {"fwup", args}
 
               _ ->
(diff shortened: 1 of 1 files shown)


mobileoverlord:
We’ve had issues in the past with using “which” to determine where the executable is. Does this not work when calling `System.find_executable?`

ConnorRigby:
I'd also prefer `System.find_executable`. The implementation is pretty much the same as `which` https://github.com/erlang/otp/blob/master/lib/kernel/src/os.erl#L167

ejpcmac:
@mobileoverlord Oh, thanks. I did not know this function. I’ll update the implementation.

mobileoverlord:
Looks good, Thanks! :bowtie:

PR #326, opened 2018-08-23
inconsistent use of MIX_TARGET=rpi3 in docs

OleMchls:
I noticed this while following this tutorial (which is great btw). But the env does not change to the next command but that runs again in `custom_rpi3` so I thought that not exporting it maybe is the right thing?

I was wondering why doing a `mix deps.get` at this point anyways. Maybe expanding the docs about that would help too.

### DIFF
--- a/docs/Systems.md
+++ b/docs/Systems.md
@@ -391,8 +391,7 @@ run the `mix nerves.system.shell` task, either from your project directory or
 from the custom system source directory.
 
 ```bash
-$ export MIX_TARGET=rpi3
-$ mix deps.get
+$ MIX_TARGET=rpi3 mix deps.get
 Mix environment
  MIX_TARGET:   rpi3
  MIX_ENV:      dev
--- a/docs/Systems.md
+++ b/docs/Systems.md
@@ -387,8 +387,7 @@ Because Buildroot can only be used from Linux, Nerves provides an abstraction
 layer called the Nerves system configuration shell that allows the same
 procedure to be used on Linux and non-Linux development hosts by using a
 Linux-based Docker container on non-Linux platforms. To access this environment,
-run the `mix nerves.system.shell` task, either from your project directory or
-from the custom system source directory.
+run the `mix nerves.system.shell` task from the custom system source directory.
 
 ```bash
 $ MIX_TARGET=rpi3 mix deps.get


ConnorRigby:
This seems like a reasonable change, since i think we use what you suggested more often anyway. 

as for _why_ we do `deps.get`
this is to fetch the deps used to actually build the system asset: `nerves`, `nerves_system_br`, `nerves_toolchain_xyz` etc.

OleMchls:
@ConnorRigby thanks for the explanation :) Do I understand you correctly that this should be executed from the `custom_rpi3` directory then? Because the docs say it does not matter? Or are the dependencies transient?

ConnorRigby:
You will need to be inside the custom system dir. `mix deps.get` will not work without being inside a mix/elixir project. (which is what custom_rpi3 is)

mobileoverlord:
@OleMchls Yes it looks like a copy / paste error. The section about configuring a custom system should be run under the context `custom_rpi3`. We should also export the target as much as possible considering it is used by more mix commands then `mix deps.*`

OleMchls:
I added a commit to specify the directory to run the mix task in. Feel free to rewrite as you like :)

docs/Systems.md
```diff
@@ -387,12 +387,10 @@ Because Buildroot can only be used from Linux, Nerves provides an abstraction
...
 procedure to be used on Linux and non-Linux development hosts by using a
 Linux-based Docker container on non-Linux platforms. To access this environment,
-run the `mix nerves.system.shell` task, either from your project directory or
-from the custom system source directory.
+run the `mix nerves.system.shell` task from the custom system source directory.
 
 ```bash
-$ export MIX_TARGET=rpi3
-$ mix deps.get
+$ MIX_TARGET=rpi3 mix deps.get
```
GregMefford: Should this also be `custom_rpi3` instead of just `rpi3` as well?

GregMefford:
Thanks a bunch for taking the time to point this out, @OleMchls! 🎈 🚀

docs/Systems.md
GregMefford: Yes sorry, I should have been more clear about that. I think the answer is yes but I haven’t had a chance to try it out myself to verify, so it’s also in the “note to self” category. :smile_sweat:

fhunleth:
I cherry-picked the fix to the paragraph about where to run the commands and dropped the MIX_TARGET change.

@GregMefford - were you working on the other changes to the instructions? If yes, those should be tracked in their own issue rather than this PR.

GregMefford:
Yep, I will open another PR.

PR #330, opened 2018-08-28
If using Elixir 1.7 then require 1.7.3

fhunleth:
This will hopefully catch people who still have 1.7.2 installed and
prevent them from hitting the runtime: false issue. Elixir 1.6 is still
supported.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -8,7 +8,7 @@ defmodule Nerves.Mixfile do
       source_url: "https://github.com/nerves-project/nerves",
       homepage_url: "http://nerves-project.org/",
       version: "1.3.0",
-      elixir: "~> 1.4",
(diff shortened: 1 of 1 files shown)


fhunleth:
This also removes support from Elixir 1.4 and 1.5. I really don't think those versions would work anyway, but let me know if they do.

ConnorRigby:
I just (re)built a legacy Nerves application that used elixir 1.4, and it mostly worked. I'd be okay with dropping support for anything older than 1.6 however.

fhunleth:
If we had an automated way to knowing that we're 1.4-safe, then I'd be ok. It seems like 1.5 supervisor specs and 1.6 DynamicSupervisors are creeping into our code base which I'm sure makes it more difficult to actually use 1.4.

PR #342, opened 2018-09-17
Run mix format on the release config

fhunleth:


fhunleth:
Circle fails even though the tests run locally and they're in an area that I didn't think should have changed. Don't have time right now to investigate. If you don't know, I'll look at it later.

PR #344, opened 2018-10-02
Use the built-in tool wslpath

jmerriweather:
Should we use wslpath instead of the regex?

### DIFF
--- a/lib/mix/nerves/utils.ex
+++ b/lib/mix/nerves/utils.ex
@@ -194,18 +194,46 @@ defmodule Mix.Nerves.Utils do
     end
   end
 
+  def has_wslpath?() do
+    System.find_executable("wslpath") != nil
+  end
+
+  def wsl_path_accessible?(file) do
+    if has_wslpath?() do
+      {_path, exitcode} = System.cmd("wslpath", ["-w", "-a", file], stderr_to_stdout: true)
+      exitcode == 0
+    else
+      false
+    end
+  end
+
   def get_wsl_paths(file) do
-    {win_path, 0} = System.cmd("cmd.exe", ["/c", "cd"])
-    win_path = String.trim(win_path) <> "\\#{file}"
+    if has_wslpath?() do
+      # Use wslpath, available from Windows 10 1803
+      # https://superuser.com/questions/1113385/convert-windows-path-for-windows-ubuntu-bash
+      # -a    force result to absolute path format
+      # -u    translate from a Windows path to a WSL path (default)
+      # -w    translate from a WSL path to a Windows path
+      # -m    translate from a WSL path to a Windows path, with ‘/’ instead of ‘\\’
 
-    drive_letter =
-      Regex.run(~r/(.*?):\\/, win_path)
-      |> Enum.at(1)
-      |> String.downcase()
+      {win_path, 0} = System.cmd("wslpath", ["-w", "-a", file])
+      {wsl_path, 0} = System.cmd("wslpath", ["-u", "-a", file])
 
-    wsl_path = "/mnt/" <> drive_letter <> "/" <> Regex.replace(~r/(.*?):\\/, win_path, "")
-    wsl_path = Regex.replace(~r/\\/, wsl_path, "/")
-    {win_path, wsl_path}
+      {String.trim(win_path), String.trim(wsl_path)}
+    else
+      # Maintain support for Windows builds before 1803
+      {win_path, 0} = System.cmd("cmd.exe", ["/c", "cd"])
+      win_path = String.trim(win_path) <> "\\#{file}"
+
+      drive_letter =
+        Regex.run(~r/(.*?):\\/, win_path)
+        |> Enum.at(1)
+        |> String.downcase()
+
+      wsl_path = "/mnt/" <> drive_letter <> "/" <> Regex.replace(~r/(.*?):\\/, win_path, "")
+      wsl_path = Regex.replace(~r/\\/, wsl_path, "/")
+      {win_path, wsl_path}
+    end
   end
 
   def set_provisioning(nil), do: :ok
--- a/lib/mix/tasks/firmware.burn.ex
+++ b/lib/mix/tasks/firmware.burn.ex
@@ -60,12 +60,17 @@ defmodule Mix.Tasks.Firmware.Burn do
       Mix.raise("Firmware for target #{target} not found at #{fw} run `mix firmware` to build")
     end
 
-    # Create a temporary .fw file that fwup.exe is able to access
     fw =
       if is_wsl?() do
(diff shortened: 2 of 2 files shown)


lib/mix/nerves/utils.ex
```diff
@@ -195,16 +195,16 @@ defmodule Mix.Nerves.Utils do
...
+    # -a    force result to absolute path format
+    # -u    translate from a Windows path to a WSL path (default)
+    # -w    translate from a WSL path to a Windows path
+    # -m    translate from a WSL path to a Windows path, with ‘/’ instead of ‘\\’
 
-    drive_letter =
-      Regex.run(~r/(.*?):\\/, win_path)
-      |> Enum.at(1)
-      |> String.downcase()
+    win_path = System.cmd("wslpath", ["-w -a", file])
```
fhunleth: Did you test this? I'd think that this line should be `{win_path, 0} = System.cmd("wslpath", ["-w", "-a", file])`.

michaelkschmidt:
Note that `wslpath` is relatively new--this will break compatibility with Creators Update and Fall Creators Update

fhunleth:
It sounds like we should hold off on changing things right now.

jmerriweather:
Yeah sorry multiple errors in this PR. I was also thinking about this and wslpath can give a "Result not representable" when the path is not available from the Windows side: `> wslpath: <filepath>: Result not representable`. 

wslpath is available from Windows 10 1803 which was released April 2018. Windows 10 1809 is about to come out, how many versions of Windows are you planning to support?

jmerriweather:
I've fixed **and tested** this branch. I've also added a feature:

- When using nerves in WSL from a path Windows can access, don't copy the firmware

If you are happy to support only the latest version (soon to be latest two versions) of Windows 10, let me know and I'll create a new PR.

fhunleth:
@jmerriweather Thanks for fixing and updating. I reopened the PR to continue discussion.

jmerriweather:
As nerves support for _Windows Subsystem for Linux_ is experimental I don't see an issue with having an expectation that users will need the latest version of Windows 10

michaelkschmidt:
Please don’t get me wrong @jmerriweather, I think this is good stuff.  I’m all for adding support for using wslpath but there needs to be a fallback for when it’s not there. 

Windows support has always been driven by Corporate users. You can’t drop support for a version of Windows that is less than a year old—it takes months and months for big IT to approve releases when they come out

jmerriweather:
@michaelkschmidt all good, it's good to talk this through and work out the best approach :)

Microsoft recently increased support for Enterprise and Education to 30 months from release. Luckily WSL isn't available in the LTSB version which has a 10 year support cycle. If full Windows support is required we'll need to provide an alternative code path until April 2020.

If the project wants to provide full support for Windows 10 I could rework the PR to fail back to the old code path?

michaelkschmidt:
I don't think we need quite that long--I'm just advocating for 1 year-ish, so I can still use it 😉 

I think `wslpath` will work with `System.find_executable()`.  Can we fall back to the existing methods if its not there?  I think you would only need to touch `get_wsl_paths(file)` and `is_wsl_path_accessible?(file)`.  The former can fall back to the existing calls, and the latter can just default to false.  This would retain the current behavior, while giving modern WSL users a much better experience

fhunleth:
I have mixed feelings about this. Maybe you two can help out.

1. Does Windows have official rules for converting paths between WSL and Windows?
2. Are those rules at risk of changing? (I'm sure anything possible, but I'm wondering if Microsoft has a note somewhere that says the mapping is a hack and not to rely on it or something like that)

What I'm getting at is that it seems like `wslpath` reduces the risk of the WSL/Windows path conversion rules changing on us. That's a good thing since it's one less thing that can go wrong.

If, however, Windows has an unambiguous set of rules for the conversion and if they're not unwieldy, would it be too much work to code them up? I assume that the existing regular expression is close and if a few unit tests were added, would that give everyone confidence in it?

I understand the arguments about Windows versions and compatibility, so don't think that I'm ignoring it.

jmerriweather:
I've tested the above commit with Windows 10 1803, @michaelkschmidt could you test this on a previous build of Windows 10?

jmerriweather:
@fhunleth I don't know of any official rules, you can actually change the root mount path of windows using `wsl.conf` as outlined here: https://blogs.msdn.microsoft.com/commandline/2018/02/07/automatically-configuring-wsl/ 

I feel that using wslpath is the safest option as Windows injects the binary it into the Linux distribution. I'm hoping my latest commit is a good compromise

lib/mix/tasks/firmware.burn.ex
```diff
@@ -63,9 +63,14 @@ defmodule Mix.Tasks.Firmware.Burn do
     # Create a temporary .fw file that fwup.exe is able to access
```
fhunleth: The "create a temporary .fw file" comment should be moved down to line 69.5.

lib/mix/nerves/utils.ex
```diff
@@ -194,18 +194,46 @@ defmodule Mix.Nerves.Utils do
     end
   end
 
-  def get_wsl_paths(file) do
-    {win_path, 0} = System.cmd("cmd.exe", ["/c", "cd"])
-    win_path = String.trim(win_path) <> "\\#{file}"
+  def is_wslpath_available?() do
```
fhunleth: There's an Elixir style to only put `is_` in functions that can be guards, so this should be named `wslpath_available?` or `has_wslpath?`

lib/mix/tasks/firmware.burn.ex
jmerriweather: ah yes, i'll move that down

lib/mix/nerves/utils.ex
jmerriweather: I wasn't aware of that. I think has_wslpath? might be a better name

jmerriweather:
My pleasure, glad we could come to a solution that works for everyone :smiley:

lib/mix/nerves/utils.ex
```diff
@@ -194,18 +194,46 @@ defmodule Mix.Nerves.Utils do
...
-    win_path = String.trim(win_path) <> "\\#{file}"
+  def has_wslpath?() do
+    System.find_executable("wslpath") != nil
+  end
 
-    drive_letter =
-      Regex.run(~r/(.*?):\\/, win_path)
-      |> Enum.at(1)
-      |> String.downcase()
+  def is_wsl_path_accessible?(file) do
```
fhunleth: One more function to remove the `is_` prefix. Fwiw, here's the logic: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Naming%20Conventions.md#is_-prefix-is_foo

Issue #347, opened 2018-10-04
Support for wslpath has broken burning firmware if path is not accessible from Windows


jmerriweather:
My [latest commit](https://github.com/nerves-project/nerves/commit/fb6fb7f20393b8be18c65c9dee092c5189f936f2) has broken burning images from WSL in the following scenario:

- Code is located in a directory inaccessible from Windows, such as `~/repo`
- Using a Windows version that supports `wslpath`

To fix the issue i propose the following:

- Make the creation of temporary firmware location explicit and remove it from [get_wsl_paths](https://github.com/nerves-project/nerves/blob/master/lib/mix/nerves/utils.ex#L210-L237)
  - get_wsl_paths will return the translated paths of the firmware
  - get_wsl_paths will return nil for a path that is inaccessible
- Copy the temporary firmware to the Windows %temp% folder. Currently the file is copied to the working directory of the current wsl process
- Fix the code path in the above scenario so that the firmware is copied to a temporary location

I've also found that `wslpath` doesn't pass through a valid wsl path, it instead returns an "Invalid argument" error. To fix this I will need to: 
- Match on the error and return the original path

fhunleth:
Thanks for looking into this. Your proposal sounds good to me.

PR #348, opened 2018-10-04
Fix issues with wslpath and temporary files

jmerriweather:
This is a work in progress for issues raised in #347


mobileoverlord:
Looks good, thanks @jmerriweather !

PR #356, opened 2018-10-16
Call mix firmware from firmware.burn task.

### LINKED ISSUE #343: Should mix firmware.burn call mix firmware
Currently `mix firmware.burn` doesn't rebuild. I recently watched this confuse two people into thinking they had installed the latest and greatest on a MicroSD card when, in fact, they hadn't rebuilt. I'm accustomed to this behavior, but I suspect that it's not expected. What do others think?

joelbyler:
Also moving original firmware.burn task into a new burn task, which only
burns firmware to an SDCard in case folks want to continue working that
way. This is being done in an effort to avoid confusion for new nerves
users as we've noticed that some users misinterpret this task and think
that it does both.  Hopefully this change will make the tasks more
intuitive based on their task names.

Was initially intended to resolve #343 but submitting as an alternative approach which retains the functionality of the old task in case its still desired by some users.

### DIFF
--- a/lib/mix/tasks/burn.ex
+++ b/lib/mix/tasks/burn.ex
@@ -1,0 +1,125 @@
+defmodule Mix.Tasks.Burn do
+  use Mix.Task
+  import Mix.Nerves.Utils
+  alias Nerves.Utils.WSL
+
+  @switches [device: :string, task: :string]
+  @aliases [d: :device, t: :task]
+
+  @shortdoc "Write a firmware image to an SDCard"
+
+  @moduledoc """
+  Writes the generated firmware image to an attached SDCard or file.
+
+  By default, this task detects attached SDCards and then invokes `fwup`
+  to overwrite the contents of the selected SDCard with the new image.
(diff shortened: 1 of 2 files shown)


fhunleth:
@mobileoverlord Could you take a look at this? You sent part of this change before Joel so there's a conflict, but the `mix burn` part of the change is still worth your review.

mobileoverlord:
Initially I thought that owning `burn` might be too broad, but, the more I think of it, the more I like how the command would look `mix do firmware, burn`. I think that this is a good idea.

mobileoverlord:
@joelbyler Could you rebase this please

joelbyler:
@mobileoverlord rebased as requested, also fwiw I've retained the `firmware.burn` task which does exactly what your previous change did, this is only to add the new `burn` task which can burn without rebuilding firmware (like the old version of `firmware.burn`).  I'm thinking some folks might prefer this at times when they're in a hurry and know that they don't need to rebuild.

mobileoverlord:
❤️ Thanks @joelbyler !

PR #360, opened 2018-10-25
Add the whole Beagle family to the targets table

smedegaard:


mobileoverlord:
Thanks for the addition!

PR #364, opened 2018-11-16
Allow http_opts to be overwritten

michaelkschmidt:
Github Enterprise seems to need `autoredirect: true` to download
large-ish packages


mobileoverlord:
Looks good to me. I imagine you are using one of the artifact site helpers. Maybe we should stuff something in its docs?

mobileoverlord:
Actually, I think I was thinking of the unsupported help text. never-mind

PR #368, opened 2018-11-19
update FAQ.md doc to reference correct rpi3 tty for USB serial console

jjcarstens:
Really simple change. I just noticed that the FAQ doc in `Using a USB Serial Console` section references example rpi3 setting to use `ttyS0` while the linked hardware description README and default `erlinit.config` both state to use `ttyAMA0`

I'm assuming it needs to be the later one, so here you go.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -25,7 +25,7 @@ For example, for the Raspberry Pi 3 target, you can find the [hardware descripti
 
     # Specify the UART port that the shell should use.
     #-c tty1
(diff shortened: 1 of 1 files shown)


ConnorRigby:
For reference: [here](https://github.com/nerves-project/nerves_system_rpi3/blob/master/rootfs_overlay/etc/erlinit.config#L8) is the current rpi3 doc.

mobileoverlord:
Thanks! :bowtie:

PR #369, opened 2018-11-21
Bump distillery to 2.0.12

mobileoverlord:
Cleans up compiler warnings.
Tested on rpi0


mobileoverlord:
Oh right I forgot about that part since they were both on the same line 😄

PR #370, opened 2018-11-21
Default strip_release: true in release config

mobileoverlord:
This will enable the setting which will work once this PR lands https://github.com/bitwalker/distillery/pull/591

Stripping the release saves some significant space.
On `rpi0`, the size of firmware for a trivial app went from `29.91 Mb` to `21.12 Mb`


GregMefford:
Does this end up stripping the docs as well?
https://elixirforum.com/t/minimizing-otp-release-size/18093/7

fhunleth:
This strips the docs. I've been running with it for a week or so, and it's a little annoying to not have the docs. On the other hand, this is in the template so it's easy enough for someone undo if they want. I'm going to hit "approve" since I'm quite used to symbols being stripped in production builds.

PR #377, opened 2019-01-24
Add support for Mix.target()

mobileoverlord:
This PR will add support for using `Mix.target()` in Elixir >= 1.8. It is also backwards compatible with < 1.8 projects.


.circleci/config.yml
```diff
@@ -1,16 +1,25 @@
...
+    command: |
+      wget https://repo.hex.pm/builds/elixir/v$ELIXIR_VERSION.zip
+      unzip -d /usr/local/elixir v$ELIXIR_VERSION.zip
+      echo 'export PATH=/usr/local/elixir/bin:$PATH' >> $BASH_ENV
+
 version: 2
 jobs:
   build:
     docker:
-      - image: circleci/elixir:1.7.3
```
fhunleth: Since this is backwards compatible, I think we should keep running CI on 1.7.3 (or 1.7.4). The mix.exs also says that 1.6 works. Do we want to keep supporting that?

mobileoverlord: I've updated the tests suite to run for the same version set as `nerves_runtime`. This ends up being a mix of otp 20 and 21 for ~> 1.6, 1.7, 1.8

fhunleth:
I made some edits for your review.

mobileoverlord:
Looks good, thanks!

PR #405, opened 2019-04-03
Adds example of overwriting files in boot partition

deanchouinard:


GregMefford:
Thank you, this is really helpful and I like how you show how to find the file locally. I usually tell people to get it from GitHub, but I bet that's more confusing for people.

PR #407, opened 2019-04-03
Bump ex_doc and other deps

fhunleth:


mobileoverlord:
looks like you updated plug (used for testing) when you updated everything, you'll have to add `plug_cowboy` to update.

fhunleth:
Yeah, I see the error. Sigh.

PR #408, opened 2019-04-05
Fix command to run Phoenix server

vtim:
`phoenix.server` Is not a valid mix task, should be `phx.server`


mobileoverlord:
Thanks!

Issue #411, opened 2019-04-09
system shell incompatible with OTP 21.3


robotmay:
### Environment

```
Erlang/OTP 21 [erts-10.3.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]
Elixir 1.8.1 (compiled with Erlang/OTP 21)
```

```
|nerves_bootstrap| Environment Package List
  Pkg:         nerves_system_br
  Vsn:         1.7.1
  Type:        system_platform
  BuildRunner: {nil, []}
  Pkg:         nerves_toolchain_armv6_rpi_linux_gnueabi
  Vsn:         1.1.0
  Type:        toolchain
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}
  Pkg:         nerves_toolchain_ctng
  Vsn:         1.5.0
  Type:        toolchain_platform
  BuildRunner: {nil, []}
  Pkg:         tyos_rpi0
  Vsn:         1.7.1
  Type:        system
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}
|nerves_bootstrap| Loadpaths Start
Nerves environment
  MIX_TARGET:   target
  MIX_ENV:      dev
|nerves_bootstrap| Environment Variable List
  target:     target
  toolchain:  /home/robotmay/.nerves/artifacts/nerves_toolchain_armv6_rpi_linux_gnueabi-linux_x86_64-1.1.0
  system:     /home/robotmay/.nerves/artifacts/custom_rpi0-portable-1.7.1
  app:        /home/robotmay/Development/elixir/test/sys/custom_rpi0
|nerves_bootstrap| Loadpaths End
```

My terminal environment is i3, alacritty, tmux, fish. I have also tried this command from inside bash and gnome terminal with the same result.

### Expected behavior

To launch the shell, hopefully.

I suspect this is something to do with i3, as it looks like it's failing to calculate a window size. I'm having quite a lot of trouble getting set up to build a custom system but I've gotten really stuck on this bit. Any ideas?

### Current behavior

```
⋊> ~/D/e/t/s/custom_rpi0 on master ⨯ mix nerves.system.shell                                                                                                                                                                                                                                                                                   15:25:05
** (ArgumentError) argument error
                                     :erlang.port_control(#Port<0.6>, 100, [])
                                                                                  lib/mix/nerves/shell.ex:75: Mix.Nerves.Shell.get_tty_geometry/1
[...]

Issue #389, opened 2019-03-11
Nerves System Shell adds current working directory $PATH, breaking linux-menuconfig


mattludwigs:
### Environment

* Elixir version: 1.8.1
* Nerves environment

```
|nerves_bootstrap| Environment Package List

  Pkg:         nerves_system_br
  Vsn:         1.6.6
  Type:        system_platform
  BuildRunner: {nil, []}

  Pkg:         nerves_system_matt
  Vsn:         1.6.2
  Type:        system
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}

  Pkg:         nerves_toolchain_arm_unknown_linux_gnueabihf
  Vsn:         1.1.0
  Type:        toolchain
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}

  Pkg:         nerves_toolchain_ctng
  Vsn:         1.5.0
  Type:        toolchain_platform
  BuildRunner: {nil, []}

|nerves_bootstrap| Loadpaths Start


Nerves environment
  MIX_TARGET:   target
  MIX_ENV:      dev

|nerves_bootstrap| Environment Variable List
  target:     target
  toolchain:  /home/mattludwigs/.nerves/artifacts/nerves_toolchain_arm_unknown_linux_gnueabihf-linux_x86_64-1.1.0
  system:     /home/mattludwigs/.nerves/artifacts/nerves_system_matt-portable-1.6.2
  app:        /home/mattludwigs/code/smartrent/nerver_system_matt

|nerves_bootstrap| Loadpaths End

```

I am using `asdf` to manage Elixir and Erlang versions and I am on Linux.

```
Nerves:           1.4.0
Nerves Bootstrap: 1.5.0
Elixir:           1.8.1
```

### Current behavior

I run `mix nerves.system.shell` and my `$PATH` gets updated with `asdf` elixir paths, but adds two `::` which resolves to the current working directory, which `make linux-menuconfig` will not work if your current working directory is in your `$PATH` 

There is this error
```
You seem to have the current working directory in your
PATH environment variable. This doesn't work.
support/dependencies/dependencies.mk:21: recipe for target 'dependencies' failed
make[1]: *** [dependencies] Error 1
Makefile:23: recipe for target '_all' failed
make: *** [_all] Error 2
```

### Expected behavior

My Path not to be touched, or at the very least the path update without adding the current working directory via a null path like `::`.

[...]

PR #413, opened 2019-04-26
System shell updates

mobileoverlord:
This PR addresses two issues.

1. Starting with OTP 21.3.0, requesting the geometry of the window through `:erlang.port_control/3` results in an argument error:
```
** (ArgumentError) argument error
:erlang.port_control(#Port<0.14>, 100, [])
lib/mix/nerves/shell.ex:75: Mix.Nerves.Shell.get_tty_geometry/1
lib/mix/nerves/shell.ex:31: Mix.Nerves.Shell.open/2
(mix) lib/mix/cli.ex:79: Mix.CLI.run_task/2
```

~This will comment this out until the issue is fixed.~
This fixes the issue by first trying the original control number and then attempting the number for OTP >= 21.3.0. Thanks @fhunleth ❤️ 

2. Starting with asdf >= 0.7.0, the `PATH` contained `::`. I was unable to identify where in the code this was happening. This PR sanitizes `::` for the system shell.

Resolves
https://github.com/nerves-project/nerves/issues/411
https://github.com/nerves-project/nerves/issues/389

### DIFF
--- a/lib/mix/nerves/shell.ex
+++ b/lib/mix/nerves/shell.ex
@@ -7,7 +7,7 @@ defmodule Mix.Nerves.Shell do
     # We use the tty_sl driver for input because it handles tty geometry and
     # streaming mode.
     stdin_port = Port.open({:spawn, "tty_sl -c -e"}, [:binary, :eof, :stream, :in])
-
+    Application.stop(:logger)
     # We run the command through the script command to emulate a pty
     cmd =
       "script -q /dev/null " <>
@@ -34,7 +34,7 @@ defmodule Mix.Nerves.Shell do
     stty sane rows #{h} cols #{w}; stty -echo
     export PS1=""; export PS2=""
     start() {
-    echo -e "\\e[17F\\e[0J\\e[1;7m\n  Preparing Nerves Shell  \\e[0m"
+    echo -e "\\e[25F\\e[0J\\e[1;7m\n  Preparing Nerves Shell  \\e[0m"
     echo -e "\\e]0;Nerves Shell\\a"
     export PS1="\\e[1;7m Nerves \\e[0;1m \\W > \\e[0m"
     export PS2="\\e[1;7m Nerves \\e[0;1m \\W ..\\e[0m"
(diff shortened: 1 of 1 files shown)


lib/mix/nerves/shell.ex
```diff
@@ -68,14 +78,20 @@ defmodule Mix.Nerves.Shell do
     end
   end
 
-  @ctrl_op_get_winsize 100
```
fhunleth: This is now 0x018b0900 bigger. See https://github.com/erlang/otp/commit/ad5822c6b1401111bbdbc5e77fe097a3f1b2b3cb

lib/mix/nerves/shell.ex
```diff
@@ -68,14 +78,20 @@ defmodule Mix.Nerves.Shell do
...
+    System.get_env("PATH")
+    |> String.replace("::", ":")
+    |> to_charlist()
+  end
 
-  defp get_tty_geometry(tty_port) do
-    geometry =
-      :erlang.port_control(tty_port, @ctrl_op_get_winsize, [])
-      |> :erlang.list_to_binary()
+  # @ctrl_op_get_winsize 100
```
fhunleth: Maybe try `0x018b0900+100` first and if that fails, try the old value of 100?

lib/mix/nerves/shell.ex
```diff
@@ -7,7 +7,7 @@ defmodule Mix.Nerves.Shell do
     # We use the tty_sl driver for input because it handles tty geometry and
     # streaming mode.
     stdin_port = Port.open({:spawn, "tty_sl -c -e"}, [:binary, :eof, :stream, :in])
-
+    Application.stop(:logger)
```
fhunleth: Is the `Application.stop(:logger)` needed?

mobileoverlord: Logger errors out when the user_dir is switched. Stopping logger prevents the console from writing logger errors to the console.

lib/mix/nerves/shell.ex
mobileoverlord: wow, good find thanks!

lib/mix/nerves/shell.ex
mobileoverlord: Stopping logger prevents it from printing error messages about `user_drv` being used.

mobileoverlord:
Updated this to include the suggested changed from @fhunleth. Tested on Ubuntu and Mac

PR #427, opened 2019-06-07
Add support for Elixir 1.9+ releases

mobileoverlord:
This commit will also make distillery an optional dependency. This means that you can still create firmware using distillery, but you would need to add {:distillery, "~> 2.0"} to your project's mix file.


lib/nerves/release.ex
```diff
@@ -0,0 +1,199 @@
...
+      |> Enum.join("\n")
+
+    build_path = Path.join([Mix.Project.build_path(), "nerves"])
+    File.mkdir_p(build_path)
+
+    Path.join(build_path, "rootfs.priorities")
+    |> File.write(priorities)
+  end
+
+  defp target_beam_files(boot_script, host_release_path, target_release_path) do
```
danielspofford: Not a blocker.

There is a lot of complexity in this function - it might be a good target for refactoring.

lib/nerves/distillery.ex
```diff
@@ -0,0 +1,90 @@
...
+        vm_args = Map.get(release.profile, :vm_args) || "rel/vm.args"
+        plugins = order_plugins(release.profile.plugins)
+
+        profile =
+          release.profile
+          |> Map.put(:dev_mode, false)
+          |> Map.put(:include_src, false)
+          |> Map.put(:include_erts, System.get_env("ERL_LIB_DIR"))
+          |> Map.put(:vm_args, vm_args)
+          |> Map.put(:plugins, plugins)
```
danielspofford: Might be a good idea to do a single `Map.merge/2` here where all of these defaults are defined in a single map.

lib/nerves/release.ex
mobileoverlord: Yeah, this one was rough. I think a lot of the added complexity is that we are between release tools and have to support both distillery and elixir 1.9+. Ill take a pass through and see if I can make it easier or add more comments.

PR #430, opened 2019-06-24
v1.4.5

mobileoverlord:


fhunleth:
I think the github doc move doesn’t need to be included, but don’t mind if you want it.

PR #431, opened 2019-07-03
Verify beam compiler version on mix firmware

mobileoverlord:


lib/mix/tasks/firmware.ex
```diff
@@ -198,4 +200,44 @@ defmodule Mix.Tasks.Firmware do
...
+        Elixir: #{elixirc.major}.#{elixirc.minor}.#{elixirc.patch}
+        OTP:    #{otpc.major}.#{otpc.minor}.#{otpc.patch}
+
+        Please use a version of Elixir that was compiled using the same major
+        version of OTP.
+
+        For example:
+
+        If your target is running OTP 22, you should use a version of Elixir
+        that was compiled using OTP 22.
```
fhunleth: How about adding this right after this sentence? I think the hardest part is knowing that Elixir is published as a precompiled binary against several OTP versions.

```
If you're using asdf to manage Elixir versions, run:

asdf install elixir #{System.version()}-otp-#{otpc.major}
asdf global elixir #{System.version()}-otp-#{otpc.major}
```

lib/mix/tasks/firmware.ex
```diff
@@ -198,4 +200,44 @@ defmodule Mix.Tasks.Firmware do
...
+    end
+  end
+
+  defp compiler_version(mod) do
+    with {:file, path} <- :code.is_loaded(mod),
+         {:ok, {_, [compile_info: compile_info]}} <- :beam_lib.chunks(path, [:compile_info]),
+         {:ok, vsn} <- Keyword.fetch(compile_info, :version) do
+      vsn
+      |> to_string()
+      |> Version.parse()
```
fhunleth: Ugh! I'm using Elixir 1.8.2-otp-22 and I got back `vsn = '7.4'` for `Kernel`'s compiler version and that isn't a semver. I'm not sure if you have a better way of doing this, but I wonder if we tack on `".0"`'s until there are 3 fields?

lib/mix/tasks/firmware.ex
```diff
@@ -198,4 +200,44 @@ defmodule Mix.Tasks.Firmware do
     less_than_elixir_19 = Nerves.elixir_version() |> Version.compare("1.9.0") == :lt
     less_than_elixir_19 && Code.ensure_loaded?(Mix.Tasks.Distillery.Release)
   end
+
+  defp compiler_check() do
+    with {:ok, otpc} <- compiler_version(:code),
+         {:ok, otp_requirement} <- Version.parse_requirement("~> #{otpc.major}.#{otpc.minor}"),
+         {:ok, elixirc} <- compiler_version(:"Elixir.Kernel") do
```
fhunleth: `Kernel` should have worked. Elixir should magically expand that to `:"Elixir.Kernel"`.

PR #438, opened 2019-07-14
Update compiler check recommendation.

mobileoverlord:
There is no linkage that we know of between the kernel compiler version, and the OTP version number. Removing this prevents the text from providing recommendations to install versions of OTP that do not exist.

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -223,8 +223,8 @@ defmodule Mix.Tasks.Firmware do
 
         If you're using asdf to manage Elixir versions, run:
 
-        asdf install elixir #{System.version()}-otp-#{otpc.major}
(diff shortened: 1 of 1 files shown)


fhunleth:
Rather than removal, we need to fix this so that’s there’s something actionable for users. In other words, if we can’t fix the problem, then what do we expect for anyone else.

fhunleth:
Fwiw, the fix is to update the print statement to report the OTP version included in the system. No need to link kernel compiler versions to OTP versions. Once that's done, the hardcoded OTP "22" in the previous line could be fixed as well.

fhunleth:
@djunod The instructions were fixed, but it was only 4 or 5 hours ago. I would assume that a page reload would fix it for you since it looks right to me.

mobileoverlord:
I Update this PR to instead recommend the use installs a version of elixir compiled using the system otp version number.

PR #439, opened 2019-07-14
correct installation procedure

shoz-f:
1)correct elixir version: 1.9.0-otp-22
2)alter .bash_profile to .bashrc to fit wider range of OS
   Ubuntu ignores ".bash_profile". ".bashrc" is OK.
   (refer to asdf-vm installation procedure)


fhunleth:
Looks good. Thanks!

PR #443, opened 2019-07-25
Add strip_beams to upgrade guide

fhunleth:


jjcarstens:
Is there a difference between `:strip_beam` and `:strip_beams`? The elixir docs seem to reference both:
https://hexdocs.pm/mix/Mix.Tasks.Escript.Build.html

fhunleth:
:scream: 

I've just been copy/pasting that line around. It definitely does something since my dev builds are massive now.

fhunleth:
It looks like singular is "right". Hmmm....

https://github.com/elixir-lang/elixir/blob/v1.9.1/lib/mix/lib/mix/tasks/escript.build.ex#L166

jjcarstens:
K, I think I found it. Looks like a typo: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/escript.build.ex#L166

fhunleth:
Jon found this, but posting here. The release option is plural: https://github.com/elixir-lang/elixir/blob/1bb106727379f9cf35a818f3716d97e0e3e46620/lib/mix/lib/mix/release.ex#L673. The singular one is for escripts.

PR #425, opened 2019-06-04
Add 9P path \\wsl$ for mix firmware.burn inside WSL on Windows 10 Version 1903

nabinno:
This PR will also enable `mix firmware.burn` for 1903 users 😊

**Reproduction:** Run `mix firmware.burn` inside WSL on Windows 10 Version 1903.
```
$ mix firmware.burn
** (Mix) Could not auto detect your SD card
```

**Reason:** With the introduction of 9P server, each distro's files can be accessed from `\\wsl$\{distro name}\` path.
```sh
$ wslpath -w -a .     
\\wsl$\Ubuntu-18.04\home\nab\hello_nerves\  
```
Cf. https://devblogs.microsoft.com/commandline/whats-new-for-wsl-in-windows-10-version-1903/

PR #455, opened 2019-10-05
Improve WSL support in preflight checks

GregMefford:
While working on #425, I noticed that WSL was requiring that fwup be installed both in WSL and in Windows in order to successfully build and burn a firmware, but it wasn't checking for both in the preflight check, resulting in unhelpful error messages.

The main change in this PR is to check for both `fwup` and `fwup.exe` executables (of the required versions) when running in WSL. I also took the opportunity to pull out the related code into a module so that there's a few less things in the `Utils` junk-drawer.

### DIFF
--- a/lib/mix/nerves/preflight.ex
+++ b/lib/mix/nerves/preflight.ex
@@ -1,28 +1,58 @@
 defmodule Mix.Nerves.Preflight do
+  alias Nerves.Utils.WSL
+
   @fwup_semver "~> 1.2.5 or ~> 1.3"
 
   def check! do
-    {_, type} = :os.type()
-    check_requirements("fwup")
-    ensure_available!("mksquashfs", package: "squashfs")
-    check_host_requirements(type)
+    :os.type()
+    |> case do
+      {_, :linux} -> if WSL.running_on_wsl?(), do: :wsl, else: :linux
+      {_, type} -> type
+    end
+    |> check_platform!()
+
     Mix.Task.run("nerves.loadpaths")
   end
 
-  defp check_requirements("fwup") do
-    ensure_available!("fwup")
+  # OSX
+  defp check_platform!(:darwin) do
+    ensure_fwup_version!()
+    ensure_available!("mksquashfs", package: "squashfs")
+    ensure_available!("gstat", package: "gstat (coreutils)")
+  end
 
-    with {vsn, 0} <- System.cmd("fwup", ["--version"]),
+  # NOTE: We currently require fwup to be installed both in WSL and in Windows
+  # because the fwup.exe in Windows is used when burning firmware to a physical
+  # memory card, and fwup in WSL is used for most other operations.
+  #
+  # WSL 2 adds some new lower-level integrations that may eliminate the need
+  # for this in the future, but we'll need to do some more research.
+  defp check_platform!(:wsl) do
+    ensure_fwup_version!()
+    ensure_fwup_version!("fwup.exe")
+    ensure_available!("mksquashfs", package: "squashfs")
+  end
+
+  # Non-WSL Linux
+  defp check_platform!(_) do
+    ensure_fwup_version!()
+    ensure_available!("mksquashfs", package: "squashfs")
+  end
+
+  defp ensure_fwup_version!(fwup_bin \\ "fwup") do
+    ensure_available!(fwup_bin)
+
+    with {vsn, 0} <- System.cmd(fwup_bin, ["--version"]),
          vsn = String.trim(vsn),
          {:ok, req} = Version.parse_requirement(@fwup_semver),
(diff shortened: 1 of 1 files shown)


lib/mix/nerves/utils.ex
```diff
@@ -14,77 +14,8 @@ defmodule Mix.Nerves.Utils do
...
-  defp missing_package_message(package) do
-    """
-    #{package} is required by the Nerves tooling.
-
-    Please see https://hexdocs.pm/nerves/installation.html#host-specific-tools
-    for installation instructions.
-    """
-  end
+  @deprecated "Use Mix.Nerves.Preflight.check!/0 instead"
+  def preflight, do: Mix.Nerves.Preflight.check!()
```
GregMefford: Do I need to have this here? I don't _think_ people should be using this API other than the other Mix tasks in this repo, which I already updated.

fhunleth:
This commit does a few things:

1. Pulls out the preflight code into its own module
2. Cleans up a few stylistic things
3. Fixes a `fwup`/`fwup.exe` check

Could you separate these out into separate PRs to make this easier to review?

GregMefford:
Sure, np.

lib/mix/nerves/utils.ex
```diff
@@ -14,77 +14,8 @@ defmodule Mix.Nerves.Utils do
...
-  defp missing_package_message(package) do
-    """
-    #{package} is required by the Nerves tooling.
-
-    Please see https://hexdocs.pm/nerves/installation.html#host-specific-tools
-    for installation instructions.
-    """
-  end
+  @deprecated "User Mix.Nerves.Preflight.check!/0 instead"
+  def preflight, do: Mix.Nerves.Preflight.check!()
```
GregMefford: Should I leave this here or just delete it? My guess is that this is an internal API that we don't expect people to be using, but it also doesn't hurt to leave it here, so I'm looking for feedback either way.

GregMefford:
I got things rebased with step-by-step commits. I tested that it's not breaking things on OSX, and I will check in WSL when I get home in a bit.

GregMefford:
Confirmed again on WSL. ✅

lib/mix/nerves/preflight.ex
```diff
@@ -38,12 +66,6 @@ defmodule Mix.Nerves.Preflight do
     end
   end
 
-  defp check_host_requirements(:darwin) do
```
fhunleth: Was this accidentally left over from the move to preflight?

fhunleth:
Thanks, though, what I meant to say is to break this up into multiple PRs so that it could be reviewed and merged incrementally.

I did start looking through the commits. The last one (documentation URL fix) looks fine. I started looking through the first few, but timed out. I didn't see anything obvious, but CircleCI didn't built them and I didn't get a chance to run the tests locally on them. I assume it builds and runs?

Also, the "Add WSL-specific Preflight checks" commit is the actual fix, right?

GregMefford:
Oh, ok. At this point, do you want me to split up to multiple PRs, or just review the commits here?

Yes, the "Add WSL-specific Preflight checks" commit is the part that actually treats WSL as a different platform and checks that both `fwup` and `fwup.exe` are available.

lib/mix/nerves/preflight.ex
GregMefford: It wasn't an accident, but yes it's being removed here because it's no longer needed since `check_platform!(:darwin)` (line 17) is already doing what this function used to do.

GregMefford:
The CircleCI build failure looked like it was just because CI was too slow and the log line it was waiting for hadn't been printed yet. When I did a rebuild, it passed in all the environments.

fhunleth:
@GregMefford I was thinking that we should do a Nerves release soon, but it seems good to get the WSL PR over the finish line. I don't think there's much left, is there?

GregMefford:
@fhunleth Correct - all the pieces are in place now and I just need to rebase this. I'll try to knock it out over lunch today.

GregMefford:
Got this rebased. Let me fire up my Windows machine and make sure it's working as expected.

GregMefford:
In WSL, starting with neither `fwup` (in WSL) or `fwup.exe` (in Windows) installed:

```
$ mix firmware
** (Mix) fwup is required by the Nerves tooling.

instructions.

$ sudo dpkg -i /home/greg/fwup_1.3.2_amd64.deb
[ ... ]

$ mix firmware

instructions.

# In an administrative command prompt on Windows side:
C:\WINDOWS\system32>choco install fwup
[ ... ]

# Back in WSL:
$ mix firmware

Nerves environment
  MIX_TARGET:   host
  MIX_ENV:      dev
```

✅ Seems to be working as expected on Windows.

mobileoverlord:
🎉

PR #461, opened 2019-10-24
Extract ensure_available!/2 preflight helper

GregMefford:
This simplifies checking for the availability of certain binaries and giving an appropriate error message saying which package to install if it's missing.

Broken out from #455


fhunleth:
@GregMefford Something happened and this one needs to be rebased. I thought it would be trivial, but there were enough conflicts that I think you should take a look.

GregMefford:
Ah yeah I was thinking that might happen. Un-indenting that big `case` statement probably confused things even though there a whitespace-only commit. I'll take a look in a bit.

PR #462, opened 2019-10-26
Move Mix task preflight checks to a module

GregMefford:
Note that the function name was change from `Mix.Nerves.Utils.preflight/0` to `Mix.Nerves.Preflight.check!/0`. This was technically a public function but I'm assuming that this isn't an API anyone is using, so I didn't leave a deprecated stub in the `Utils` module. Let me know if you prefer for it to stay there.

Split out from #455


GregMefford:
Any reason not to merge this one?

Issue #386, opened 2019-02-22
nerves: follow XDG base dir specification


fhunleth:
Nerves stores global data to `~/.nerves`. The [XDG Base Dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) provides a standard location for these kinds of files. There's an Elixir issue and discussion that have some compelling reasons behind changing. It seems like Nerves should change as well. See:

* https://github.com/elixir-lang/elixir/issues/8818
* https://groups.google.com/forum/#!topic/elixir-lang-core/7nyDfkgRsI0

mobileoverlord:
Starting in Nerves [1.5.2](https://github.com/nerves-project/nerves/releases/tag/v1.5.2) 
 the data directory for Nerves will use `XDG_DATA_HOME` if set. We will revisit supporting the XDG spec more closely in a major version upgrade.

PR #463, opened 2019-10-29
Add NERVES_DATA_DIR to specify global nerves data location

mobileoverlord:
This also adds support for XDG_DATA_HOME as described in  https://github.com/elixir-lang/elixir/pull/8933

This is going to require some documentation but I wanted to start the discussion on the variable name. Considering that the subcomponents of this could be set via `NERVES_DL_DIR` and `NERVES_ARTIFACTS_DIR` I chose to fix the existing but broken `data_dir/0` function and call it `NERVES_DATA_DIR`.

### DIFF
--- a/lib/nerves/artifact.ex
+++ b/lib/nerves/artifact.ex
@@ -6,7 +6,7 @@ defmodule Nerves.Artifact do
   """
   alias Nerves.Artifact.{Cache, BuildRunners, Resolvers}
 
-  @base_dir Path.expand("~/.nerves/artifacts")
+  @path Path.expand("artifacts")
   @checksum_short 7
 
   @doc """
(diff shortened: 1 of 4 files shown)


mobileoverlord:
Fixes https://github.com/nerves-project/nerves/issues/386

fhunleth:
@mobileoverlord What do you think about this line in the XDG spec:

>$XDG_DATA_HOME defines the base directory relative to which user specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used. 

That reads to me that defaulting to `~/.nerves` if `$XDG_DATA_HOME` is not set is not the intended behavior.

mobileoverlord:
If we change the default from `~/.nerves` to `$HOME/.local/share` it would happen for everyone, which should be okay. I wonder if there is another `XDG` variable that is guaranteed to be set and could be used to determine which default to choose?

fhunleth:
Yes, that would be confusing. It sounds like completely following XDG is a Nerves 2.0 thing.

PR #464, opened 2019-11-08
Allow erlinit options to be merged from mix config

mobileoverlord:
The options from the system are merged with the options specified in `Application.get_env(:nerves, :erlinit)`. If there are overrides, a new erlinit.config file is generated and placed in a new _build  rootfs_overlay directory. This path is passed to rel2fw and will be overridden by the project level rootfs_overlay files.

Here is an example:
If we have a Raspberry PI 3 and in our `config/target.exs` put 

```elixir
config :nerves, :erlinit,
  ctty: "ttyAMA0"
```
The generated config will send the terminal output to the serial port.

There are a few caveats with this approach.

  * Options that accumulate can not be overridden, they can only be appended. For example, multiple `--mount`.
  * Other options can be removed if a value is already specified by setting the value to `nil` or `false` in the case of `boolean` values. For example, disabling the `--alternate_exec` `alternate_exec: nil` or disabling `--verbose` `verbose: false`.

### DIFF
--- a/docs/Advanced Configuration.md
+++ b/docs/Advanced Configuration.md
@@ -395,3 +395,54 @@ attempt to mount the partition. You may want to see [how `nerves_runtime` does
 this for the default application data
 partition](https://github.com/nerves-project/nerves_runtime/blob/master/lib/nerves_runtime/init.ex),
 extending it to meet your specific needs.
+
+### Overriding erlinit.config from Mix Config
+
+Options specified in the `erlinit.config` file can be overridden through the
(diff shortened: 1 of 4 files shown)


ConnorRigby:
yes yes yes yes yes

fhunleth:
Don't hate me. I know this is a huge pain.

I'd like to sit on this for a week or two so that we can work through how we want to implement automatic recovery on rpi and x86. If that change involves updating erlinit, we might solve this another way.

PR #467, opened 2019-11-18
Allow source_date_epoch to be set from Config

mobileoverlord:
This will set `SOURCE_DATE_EPOCH` to a value from `config :nerves, source_date_epoch: "1234"` or can be overridden if `SOURCE_DATE_EPOCH` is already defined.

### DIFF
--- a/lib/nerves/env.ex
+++ b/lib/nerves/env.ex
@@ -17,6 +17,7 @@ defmodule Nerves.Env do
   """
   @spec start() :: Agent.on_start()
   def start do
+    set_source_date_epoch()
     Agent.start_link(fn -> load_packages() end, name: __MODULE__)
   end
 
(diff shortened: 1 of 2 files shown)


lib/nerves/env.ex
```diff
@@ -386,6 +387,21 @@ defmodule Nerves.Env do
...
+      System.put_env("SOURCE_DATE_EPOCH", source_date_epoch)
+    end
+  end
+
+  defp source_date_epoch() do
+    (System.get_env("SOURCE_DATE_EPOCH") || Application.get_env(:nerves, :source_date_epoch))
+    |> case do
+      nil -> nil
+      int when is_integer(int) -> Integer.to_string(int)
+      str when is_binary(str) -> str
```
fhunleth: Could we have a check on the string to make sure that it's an integer? Specifically, I'd like to protect against someone setting `SOURCE_DATE_EPOCH` to the empty string and getting 1970 through out. I may have made that mistake.

fhunleth:
I like it. I'm pretty sure that I'd enabled this on every project I make.

mobileoverlord:
I was going to add this to the new project generator once we get it in.

lib/nerves/env.ex
mobileoverlord: I updated this to validate the input

PR #468, opened 2019-11-19
v1.5.2

mobileoverlord:


CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## v1.5.2
+
+* Enhancements
+  * Added `NERVES_DATA_DIR` environment variable for specifying the location
+    for global artifacts and downloads.
+  * Configure `NERVES_DATA_DIR` to `XDG_DATA_HOME` if defined.
+  * Add `nerves` config for `source_date_epoc` to enable reproducible builds.
```
fhunleth: * Nerves tooling now supports setting the `SOURCE_DATE_EPOCH` environment variable for reproducible builds during compilation via `:source_date_epoch` in your application config. This removes timestamp differences between builds. See [reproducible-builds.org](https://reproducible-builds.org/) for more information.

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+  * Allow `erlinit.config` options to be overridden from Mix config.
+    The options from the system are merged with the options specified in the
+    config for `:nerves, :erlinit`.
```
fhunleth: * `erlinit.config` options can be overridden using the application config now. For example, in your `config.exs` you can now add:

```
config :nerves, :erlinit,
  ctty: "ttyAMA0"
```

I think we need to update the docs for this feature. I couldn't find any documentation except for the source code. Here's what we currently say: https://github.com/nerves-project/nerves/blob/553de3a139402ed47b7ba307eb1379ab017f0225/docs/Advanced%20Configuration.md#overwriting-files-in-the-root-filesystem

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## v1.5.2
+
+* Enhancements
+  * Added `NERVES_DATA_DIR` environment variable for specifying the location
```
fhunleth: I think this one is just:

  * Support `XDG_DATA_HOME`. If `XDG_DATA_HOME` is set, Nerves will now store its data under that directory.

This feels like the least important enhancement, so my vote would be to list it last. The `erlinit` one feels like the most important one or at least the one that's caused the most annoyance.

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+    for global artifacts and downloads.
+  * Configure `NERVES_DATA_DIR` to `XDG_DATA_HOME` if defined.
+  * Add `nerves` config for `source_date_epoc` to enable reproducible builds.
+  * Allow `erlinit.config` options to be overridden from Mix config.
+    The options from the system are merged with the options specified in the
+    config for `:nerves, :erlinit`.
+
+* Bug fixes
+  * Improved support for identifying Windows Subsystem for Linux in mix task
```
fhunleth: This seems like an improvement rather than a bug fix.
  * Windows Subsystem for Linux improvements

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+    preflight checks.
+  * Improved error messages when missing host packages.
```
fhunleth: Skip this one.

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+  * Do not require sudo on `mix burn` if already privileged.
+  * Keep start and start_clean boot scripts when making a release.
```
fhunleth: * Keep all boot scripts. Previously, extraneous boot scripts from the OTP release process were removed. Keeping them makes it possible to start Erlang slave nodes and support use cases where triggers at device boot time launch different scripts.

CHANGELOG.md
mobileoverlord: I added a blurb about it here https://github.com/nerves-project/nerves/blob/553de3a139402ed47b7ba307eb1379ab017f0225/docs/Advanced%20Configuration.md#overriding-erlinitconfig-from-mix-config

Maybe we need to move it above the section mentioned?

PR #474, opened 2019-12-16
Allow the path to the fw to be passed to mix burn

mobileoverlord:
This will update the task to allow calling `mix burn --firmware /path/to/app.fw`


fhunleth:
Isn't it faster and easier to call `fwup /path/to/app.fw`?

mobileoverlord:
> Isn't it faster and easier to call `fwup /path/to/app.fw`

Yes. This PR is being staged for continuity between mix tasks. The desire was uncovered through calls to `mix nerves_hub.device provision` and pointing to an existing library of fw files. That task will call through to `mix burn`. Another PR to handle this is incoming.

PR #477, opened 2019-12-27
Add WSL installation instructions

tonnenpinguin:
I've had a hard time getting nerves to build under WSL 2 as it kept showing the following error message when running `mix firmware`:
```
instructions.
```

This PR adds a short section to the installation instructions that should help people with more recent WSL releases to figure out the right setup right away :)

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -39,6 +39,20 @@ faster you can compile a custom Nerves system.
 
 Now skip to the instructions for all platforms below.
 
+## Windows Subsystem for Linux (WSL)
+Make sure that your `/etc/wsl.conf` file looks something like the following. If the file does not exist yet, create it.
+
+```
+[interop]
+enabled=true # enable launch of Windows binaries
+appendWindowsPath=true # append Windows path to $PATH variable
+```
+
+If you changed the WSL config don't forget to reboot the WSL by running  `wsl --shutdown` in PowerShell and starting the WSL shell again.
+
+Now follow the Linux instructions below.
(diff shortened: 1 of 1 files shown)


fhunleth:
@tonnenpinguin Thanks for sending this!

@GregMefford Does this look good to you?

GregMefford:
Sorry this sat around for so long, I missed it when it happened. I'll give it a try on my Windows machine tonight and verify it's working properly for me, and we can get it merged!

GregMefford:
After far more effort than I expected across multiple evenings and more than one computer, I finally got WSL 2 working, haha! 😅 

@tonnenpinguin: I wasn’t able to reproduce the behavior you’re describing in this PR. It looks like the configuration you’re recommending is the same as the defaults, which appears to be how it’s working on my mostly-fresh Windows machine.

https://devblogs.microsoft.com/commandline/automatically-configuring-wsl/

If you delete your `/etc/wsl.conf` completely (or just remove the `interop` settings mentioned in this PR), are you still able to run `fwup.exe`? If so, then I think that no additional steps need to be documented here because it should just work unless you changed the defaults yourself.

fhunleth:
I'm going to close this since it seems like everything is ok. Please re-open or create a new issue if not.

tonnenpinguin:
@GregMefford Sorry for the long silence.

I've been using my windows install for a while now, so it might actually be something related to systems upgraded over time rather than fresh installs.

At least google can now be a bit more helpful if somebody is running into the same issue I did :)

PR #479, opened 2020-01-10
Fix Dialyzer warnings

fhunleth:
This fixes a few of the (sadly) many Dialyzer warnings.

The only nontrivial one is the version parsing. Since `Version.parse/1`
doesn't return an error tuple (it just returns `:error`), Dialyzer would
complain that a lone `:error` atom would need to be handled as well. It
looks like the regular expressions should prevent errors from being
returned in practice, so I changed the parsing to raise should the
regular expressions miss anything.


mobileoverlord:
Switching to a bang scares me a little, the context of it might be safe now, but do we have that guarantee for the future? Should we match on the return and wrap it? I could be overthinking this, is it really trivial?

fhunleth:
I'm pretty sure that the code is guaranteed to work and if those bangs ever raised, we'd want to understand why. At least, it passed my threshold on when to raise vs return an error tuple. I.e., this feels more like programmer error if it ever raises and not a consequence of bad input.

PR #485, opened 2020-01-26
Update erlinit config path for system env var

mobileoverlord:
If NERVES_SYSTEM is set, look for the erlinit.config file in the system override location instead of the package path.


fhunleth:
Whoa... I'm surprised that I didn't run into this!

PR #490, opened 2020-02-28
Support the aarch64 architecture

fhunleth:
64-bit ARM machines previously were falling through the x86_64-catchall.
This adds an entry for them since they'll likely appear more often going
forward.


fhunleth:
@zacky1972 - This is one part of being able to build Nerves firmware on 64-bit ARM processors. We'll still need toolchains, so there's more work to do.

PR #498, opened 2020-03-18
Add upgrade instructions for Nerves 1.6

fhunleth:


docs/Updating Projects.md
```diff
@@ -735,3 +736,28 @@ sections of the `config.exs`. Here's one way of doing this:
       import_config "target.exs"
     end
     ```
+
+## Updating from v1.5 to v1.6
+
+Nerves 1.6 adds support for Elixir 1.10. In truth, only the internals of the
+Nerves tooling were changed. As a result of this change, we made the decision to
+drop support for Elixir 1.6 and Erlang 20. If you are still using an old version
+update to supported versions first. Then update to Nerves 1.6.
```
jjcarstens: Could just be me, but this is slightly confusing here. Are we talking about elixir and Erlang being the "old" versions? And if so, "update to supported versions" would mean get to Elixir 1.9 OTP 22 Nerves 1.5, get things working, then update elixir 1.10 and Nerves 1.6?

fhunleth: Do you have suggested text?

jjcarstens:
```suggestion
drop support for Elixir 1.6 and Erlang 20. If you are still using the older
versions, you'll need to update to at least Elixir 1.7 and Erlang 21. Then
update to Nerves 1.6.
```

fhunleth: I like it.

PR #499, opened 2020-03-18
Require fwup 1.5 or later

fhunleth:
While most users won't need 1.5, updating to it brings in fixes and
includes GPT and SOURCE_DATA_EPOCH support which will be nice to expect.

Fwup 1.5 was released November 16, 2019.


jjcarstens:
Change is fine with me. Not sure why Elixir 1.10 would be failing though

PR #504, opened 2020-03-25
Bump deps to latest versions

mobileoverlord:


mix.exs
```diff
@@ -29,11 +29,11 @@ defmodule Nerves.Mixfile do
   defp deps do
     [
       {:distillery, "~> 2.1", optional: true, runtime: false},
-      {:jason, "~> 1.0", optional: true},
-      {:ex_doc, "~> 0.19", only: [:test, :dev], runtime: false},
```
fhunleth: `ex_doc v0.20` fails on Elixir versions before 1.8 or 1.9, right?

mobileoverlord: I remember this too. I just tested on Elixir 1.7.4
```
Docs successfully generated.
View them at "doc/index.html".
➜  nerves git:(bump-deps) ✗ elixir --version
Erlang/OTP 22 [erts-10.6.4] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe]

Elixir 1.7.4 (compiled with Erlang/OTP 22)
```

fhunleth: huh

PR #505, opened 2020-04-02
Add pkg-config to installation instructions

mobileoverlord:
Since switching to `nerves_pack`,`vintage_net_wifi` is included by default and requires pkg-config to be installed on the host system.

Fixes:
```
==> vintage_net_wifi
Makefile:53: *** pkg-config required to build. Install by running "brew install pkg-config".  Stop.
could not compile dependency :vintage_net_wifi, "mix compile" failed. You can recompile this dependency with "mix deps.compile vintage_net_wifi", update it with "mix deps.update vintage_net_wifi" or clean it with "mix deps.clean vintage_net_wifi"
```

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -25,7 +25,7 @@ Just run the following:
 
 ```bash
 brew update
-brew install fwup squashfs coreutils xz
+brew install fwup squashfs coreutils xz pkg-config
 ```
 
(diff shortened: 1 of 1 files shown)


fhunleth:
Ahhhh. The reason for pkg-config is soooo annoyingly minor. I tried to not need that dependency.

PR #514, opened 2020-04-17
Update sponsor images

jjcarstens:
nerves-project.org changed which means the assets for these images changed.

So this updates them and fixes a couple other links as well


fhunleth:
Thanks for the catch!

PR #522, opened 2020-05-06
Fix Github API key validation

mobileoverlord:
I'm sorry, I screwed it up.


test/nerves/artifact/resolver_test.exs
```diff
@@ -120,6 +120,15 @@ defmodule Nerves.Artifact.ResolverTest do
     end)
   end
 
+  test "github api validate keys" do
```
fhunleth, line 123:
```suggestion
  test "github api returns error when missing required options" do
```

PR #526, opened 2020-06-14
minor updates to docs

petrus-jvrensburg:
Update the docs with some things that I bumped my head against.


fhunleth:
Thanks - all of your updates look good!

PR #528, opened 2020-06-22
Clean the release directory

mobileoverlord:
This will clear out the release directory and prevent it from accumulating upgraded libraries / unnecessary OTP applications over time.

### DIFF
--- a/lib/nerves/release.ex
+++ b/lib/nerves/release.ex
@@ -12,6 +12,8 @@ defmodule Nerves.Release do
         steps: release.steps ++ [&Nerves.Release.finalize/1]
     }
 
+    File.rm_rf!(release.path)
+
     if Code.ensure_loaded?(Shoehorn.Release) do
       apply(Shoehorn.Release, :init, [release])
     else


lib/nerves/release.ex
```diff
@@ -12,6 +12,8 @@ defmodule Nerves.Release do
         steps: release.steps ++ [&Nerves.Release.finalize/1]
     }
 
+    File.rm_rf(release.path)
```
fhunleth, line 15: Could this use bang version, `File.rm_rf!/1`, since the return value is ignored (this should be a Dialyzer warning?) and an error would be pretty surprising, I'd think.
 
```suggestion
    File.rm_rf!(release.path)
```

mobileoverlord: The first time running the release it would fail. We would have to check if it exists and then remove. That is why I chose the version without the bang. Should we match the return or add the additional logic?

fhunleth: It shouldn't fail:

```elixir
iex(2)> File.rm_rf!("bogus/path/foo")
[]
```

PR #532, opened 2020-08-06
Vendor MuonTrap to prevent zombie processes

mobileoverlord:
This will add a port process using code copied in from MuonTrap.
All calls that were made to System.cmd/3 are being routed through this port to ensure that the child processes are terminated when the VM exits. This will resolve issues with zombie processes being created when executing build runners.


ConnorRigby:
Looks good to me. CI seems to be failing, and i don't fully understand why but when that's taken care of it gets a :+1: from me

fhunleth:
Nice. I'm going to try to integrate this into my workflow, so I can see how it works in practice. Feels like a pretty good improvement, though.

mobileoverlord:
I fixed up the issue with the tests failing and as a bonus I figured out what was causing that test in `env_test.exs` to fail randomly and I fixed it. This should be all good now.

PR #537, opened 2020-09-09
Add support for creating firmware patches in mix tasks

mobileoverlord:
This PR adds the mix task `mix firmware.patch` for testing patchable firmware files. I've added a bit of documentation about it under an "Experimental features" section since the system requirements and workflow for creating firmware patches may change.

### DIFF
--- a/docs/Experimental Features.md
+++ b/docs/Experimental Features.md
@@ -1,0 +1,217 @@
+# Experimental features
+
+The features described in this document are experimental. They are under
+consideration and or actively being developed.
+
+## Firmware patches
+
+Firmware update files (`.fw`) contain everything your target needs to boot and
+run your application. Commonly, this single file package will contain your root
+filesystem, the linux kernel, a bootloader, and some extra files and metadata
+specific to your target. Packaging all these files together provides a convenient
+and reliable means of distributing firmware that can be used to boot new devices
+as well as upgrade existing ones. Unfortunately, this mechanism is not conducive
+to applying updates to devices that are use expensive metered network connections
+where the cost of every byte counts. This problem can be alleviated with firmware
+patches.
+
+A firmware patch file's content structure is identical to that of a regular
+firmware update file, it contains your root file system, the linux kernel, and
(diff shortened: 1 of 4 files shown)


docs/Experimental Features.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Firmware patches
+
+Firmware update files (`.fw`) contain everything your target needs to boot and
+run your application. Commonly, this single file package will contain your root
+filesystem, the linux kernel, a bootloader, and some extra files and metadata
+specific to your target. Packaging all these files together provides a convenient
+and reliable means of distributing firmware that can be used to boot new devices
+as well as upgrade existing ones. Unfortunately, this mechanism is not conducive
+to applying updates to devices that are use expensive metered network connections
```
fhunleth, line 14:
```suggestion
to applying updates to devices that use expensive metered network connections
```

docs/Experimental Features.md
```diff
@@ -0,0 +1,217 @@
...
+
+```
+mix upload --firmware /path/to/test_patch/_build/rpi0_dev/nerves/images/patch.fw
+```
+
+The size difference between the `Target` output firmware size `32053263` and the
+patched firmware size `4421811` has a pretty significant size reduction. For
+such a small change, we might expect more. A lot of this size come from the
+files that are also included in the firmware that are not currently being patched
+such as the linux kernel and other files that do not change frequently.
```
fhunleth, line 214: I think this is a true statement especially on the Raspberry Pi where there are a lot of bytes trapped in the boot partition. Can this be tried with `MIX_ENV=prod`, though? It doesn't seem like anyone would be pushing dev builds to remote devices. Of course, I do development with `MIX_ENV=prod` since firmware updates are faster and I don't use `h/1` much on device.

fhunleth (approved):
This is really exciting!!!! I left a couple nits, but nothing substantial.

docs/Experimental Features.md
mobileoverlord: I updated this with `prod` env stats.

PR #538, opened 2020-09-23
add Experimental Features to docs when publishing

jjcarstens:
We have this doc, but it doesn't show on hexdocs.pm

Is this on purpose? If not, here's the change. Once merged we can do `mix hex.publish docs` to just push this up for use


mobileoverlord (approved):
Doh. I forgot to add it to the list. I was preparing for a release on this repo.

PR #543, opened 2020-09-26
Remove distillery and only support Elixir releases

mobileoverlord:
This will remove all the code that switches between distillery and Elixir releases. With Elixir 1.11 on the horizon, I think its time we support a single upstream path. I'd expect that older projects have either upgraded to Elixir releases or they could lock to a <= 1.6 version of `nerves`.


mix.exs
```diff
@@ -85,7 +84,6 @@ defmodule Nerves.MixProject do
         "mix.exs",
         "README.md",
         "scripts",
-        "priv/templates",
```
fhunleth, line 88: Wait.... the whole source priv directory fiasco yak shaved all the way to removing Distillery!?!??! 🤣

jjcarstens (approved):
I'm totally on board with this. Seems like a reasonable assertion to make and cleans up a lot. But still being in the new end of Nerves, I may be missing some other potential ramifications so here's my "approve", but someone else should prob weigh in as well

mix.exs
mobileoverlord: YOLO!!!!

fhunleth: wow

fhunleth (approved):
Looks good!

PR #545, opened 2020-09-29
Fix rootfs_priorities file creation and add test

mobileoverlord:
This fixes rootfs priority file creation logic that was broken when removing distillery, adds a end to end test.


fhunleth (approved):
Looks good and it fixes the build for me.

fhunleth:
Oh, I see that CI is failing. Count my approval after whatever is doing that is fixed.

PR #549, opened 2020-10-05
Reduce prints from Make

fhunleth:
Even when nothing changes on the C side, `make` always outputs 4 lines
due to the use of a recursive `Makefile`. This moves `src/Makefile` into
the main directory so that it no longer gets called recursively and
therefore, the prints are no longer generated.

Before:

```
$ mix firmware
==> nerves
/usr/bin/make -C src all
make[1]: Entering directory '/home/fhunleth/nerves/nerves/src'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/fhunleth/nerves/nerves/src'
...
```

After:

```
$ mix firmware
==> nerves
...
```


mobileoverlord (approved):
This is nice, thanks!

jjcarstens:
Bless your soul, Frank...

PR #551, opened 2020-10-06
Don't silently drop unknown erlinit options

fhunleth:
OptionParser silently drops options that it doesn't know. This is a
"feature" since options are turned into atoms and the Erlang VM doesn't
GC atoms. However, this is not a feature for `erlinit` since it prevents
possibly valid options from making their way to the target. Exhausting
the atom pool is not an issue, since this is a compilation tool and not
a long-lived server.


ConnorRigby (approved):
I'm glad I don't know your pain right now that caused you to find this.

jjcarstens (approved):

mobileoverlord (approved):

PR #554, opened 2020-10-06
v1.7.0 release

mobileoverlord:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## v1.7.0
+
+Nerves 1.7.0 removes support for creating OTP releases using Distillery and
+only supports using Elixir releases which bumps the minimum supported version
+of Elixir to 1.9.
```
fhunleth, line 7:
```suggestion
only supports using Elixir releases. As a result, the minimum supported version
of Elixir is now version 1.9.
```

fhunleth (approved):

fhunleth (changes requested):
Could we update the fwup check? The current check is for `~> 1.5` which will start failing as people try out delta updates. Delta updates require 1.6. See #555

fhunleth (approved):

fhunleth:
@mobileoverlord Fwiw, I first bumped the fwup requirement up to 1.6 since that's the minimum for delta updates, but then I figured that I might as well bump to 1.8 to force people to download 1.8.1 and get the MicroSD card detection improvements. I'm selfishly trying to reduce my need to support some users. Anyway, that's it.

PR #556, opened 2020-10-09
Require Elixir 1.9 or later (from Nerves v1.7.0)

takasehideki:
According to the release note, the minimum supported version of Elixir is now version 1.9.
And also, WSL "2" is recommended when we use Windows as the host.


mobileoverlord (approved):
Awesome thanks!

PR #564, opened 2020-11-16
Fix a documentation error in `docs/User Interfaces.md`

kentaro:
I removed trailing comma in a sample code to prevent syntax error when copy-and-pasted.


mobileoverlord:
Thanks 😁

PR #567, opened 2020-12-01
Run 'make source all' to catch download issues first

fhunleth:
If a source code repository or the URL to it is broken, this change will
catch that error at the beginning of the build. This saves quite a bit
of time waiting for a build to complete only to find out that one of the
last packages to build fails due to a download issue.


jjcarstens (approved):
🙏 bless your soul!

ConnorRigby (approved):

mobileoverlord (approved):

PR #569, opened 2020-12-05
check whether the env is Docker or WSL 2 shell

### LINKED ISSUE #568: fwup".exe" is required even in the Docker environment on Windows
_NOTE: This issue was appeared when we use Docker Desktop for Windows as the development environment.
I understand this issue is too unique. But don't worry, I have already found an idea to fix it._

### Environment

* Elixir version (`elixir -v`):
```
$ elixir -v 
Erlang/OTP 23 [erts-11.1.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.11.2 (compiled with Erlang/OTP 23)
```

* Nerves environment: (`mix nerves.env --info`)
```
$ mix nerves.env --info 
==> nerves
==> hello_nerves
|nerves_bootstrap| Environment Package List

  Pkg:         nerves_system_br
  Vsn:         1.13.5
  Type:        system_platform
  BuildRunner: {nil, []}

  Pkg:         nerves_system_rpi3
  Vsn:         1.13.2
  Type:        system
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}

  Pkg:         nerves_toolchain_arm_unknown_linux_gnueabihf
  Vsn:         1.3.2
  Type:        toolchain
  BuildRunner: {Nerves.Artifact.BuildRunners.Local, []}

  Pkg:         nerves_toolchain_ctng
  Vsn:         1.7.2
  Type:        toolchain_platform
  BuildRunner: {nil, []}

|nerves_bootstrap| Loadpaths Start


Nerves environment
  MIX_TARGET:   rpi3
  MIX_ENV:      dev

|nerves_bootstrap| Environment Variable List
  target:     rpi3
  toolchain:  /root/.nerves/artifacts/nerves_toolchain_arm_unknown_linux_gnueabihf-linux_x86_64-1.3.2
  system:     /root/.nerves/artifacts/nerves_system_rpi3-portable-1.13.2
  app:        /workspaces/nerves-devcontainer/hello_nerves

|nerves_bootstrap| Loadpaths End
```

* Additional information about your host, target hardware or environment that
  may help

- Docker Desktop for Windows as development environment (please also see [my post on Elixir Forum](https://elixirforum.com/t/nerves-development-environment-with-docker-and-vs-code/35973)
- WSL 2 Linux kernel was installed

### Current behavior

```
$ ls -l /usr/bin/fwup*
-rwxr-xr-x 1 root root 703000 Oct 31 16:24 /usr/bin/fwup
lrwxrwxrwx 1 root root      4 Nov 26 07:18 /usr/bin/fwup.exe -> fwup
$ rm /usr/bin/fwup.exe 
$ mix firmware 
==> nerves
==> hello_nerves

Nerves environment
  MIX_TARGET:   rpi3
  MIX_ENV:      dev

instructions.
```

### Expected behavior

The firmware build succeeds without fwup**.exe**.

### Why it happens

The current implementation checks whether the environment is WSL 2 by the kernel name (`/proc/sys/kernel/osrelease`). 

https://github.com/nerves-project/nerves/blob/5bd888800c97c126476b3ef8059b3dbf9c69d45d/lib/nerves/utils/wsl.ex#L22-L34

If it was true, `fwup.exe` is adopted as the name of fwup.

https://github.com/nerves-project/nerves/blob/5bd888800c97c126476b3ef8059b3dbf9c69d45d/lib/mix/nerves/preflight.ex#L24-L34

The reason why this issue appeared is the Docker Desktop for Windows uses WSL 2 kernel as the back-end. So this environment also has "microsoft".

```
iex(1)> System.cmd("cat", ["/proc/sys/kernel/osrelease"]) 
{"4.19.128-microsoft-standard\n", 0}
```

The simplest way to avoid this issue is to create the symbolic link of `fwup.exe` to `fwup`.
[Our Docker image](https://github.com/NervesJP/docker-nerves/blob/ae92d7aad5463992f40ed84bca5a29bc1cc09cd7/Dockerfile#L24) has created this as the follow.

```
RUN ln -s fwup /usr/bin/fwup.exe
```

However, we can evolve the code to do a smarter way.

takasehideki:
Docker Desktop for Windows uses WSL 2 kernel as the back-end. 

`/.Dockerenv` exists in the environment where Docker container is running.
We can use this file to check whether the environment is Docker or WSL 2 shell before checking the kernel name from `/proc/sys/kernel/osrelease`. 

This PR will fix #568

### DIFF
--- a/lib/nerves/utils/wsl.ex
+++ b/lib/nerves/utils/wsl.ex
@@ -22,19 +22,17 @@ defmodule Nerves.Utils.WSL do
   @doc """
   Returns true if inside a WSL shell environment
   """
-  @spec running_on_wsl?(String.t()) :: boolean
-  def running_on_wsl?(osrelease_path \\ "/proc/sys/kernel/osrelease") do
-    with true <- File.exists?(osrelease_path),
-         {content, _} <- Nerves.Port.cmd("cat", [osrelease_path]) do
-      # Docker Desktop for Windows uses WSL 2 kernel as the back-end
-      # This line checks whether the env is Docker or WSL 2 shell
-      case !File.exists?("/.dockerenv") do
-        true -> Regex.match?(~r/[Mm]icrosoft/, content)
-        false -> false
-      end
-    else
-      _ ->
-        false
+  @spec running_on_wsl?() :: boolean
+  def running_on_wsl?() do
+    # Docker Desktop for Windows uses WSL 2 kernel as the back-end
+    # so also check whether the env is Docker
+    Regex.match?(~r/[Mm]icrosoft/, osrelease()) and not File.exists?("/.dockerenv")
+  end
+
+  defp osrelease() do
+    case File.read("/proc/sys/kernel/osrelease") do
+      {:ok, text} -> text
+      _ -> "unknown"
     end
   end
 
--- a/lib/nerves/utils/wsl.ex
+++ b/lib/nerves/utils/wsl.ex
@@ -26,7 +26,12 @@ defmodule Nerves.Utils.WSL do
   def running_on_wsl?(osrelease_path \\ "/proc/sys/kernel/osrelease") do
     with true <- File.exists?(osrelease_path),
          {content, _} <- Nerves.Port.cmd("cat", [osrelease_path]) do
-      Regex.match?(~r/[Mm]icrosoft/, content)
+      # Docker Desktop for Windows uses WSL 2 kernel as the back-end
+      # This line checks whether the env is Docker or WSL 2 shell
+      case !File.exists?("/.dockerenv") do
+        true -> Regex.match?(~r/[Mm]icrosoft/, content)
+        false -> false
+      end
     else
       _ ->
         false


takasehideki:
This PR has been checked by changing `mix.exs`

```
  defp deps do
    [
      # Dependencies for all targets
      {:nerves, github: "takasehideki/nerves", braanch: "docker_with_wsl", runtime: false, override: true},
```

I've check this PR works well (without affecting others) on following environments
- WSL 2 shell
- Docker container with Docker Desktop for Windows
- macOS terminal
- Docker container with Docker Desktop for Mac

lib/nerves/utils/wsl.ex
```diff
@@ -26,7 +26,12 @@ defmodule Nerves.Utils.WSL do
...
-      Regex.match?(~r/[Mm]icrosoft/, content)
+      # Docker Desktop for Windows uses WSL 2 kernel as the back-end
+      # This line checks whether the env is Docker or WSL 2 shell
+      case !File.exists?("/.dockerenv") do
+        true -> Regex.match?(~r/[Mm]icrosoft/, content)
+        false -> false
+      end
     else
       _ ->
         false
```
fhunleth, line 37: It looks like this code could be simplified. Would something like this work?

```elixir
  def running_on_wsl?() do
    # Docker Desktop for Windows uses WSL 2 kernel as the back-end
    # so also check whether the env is not Docker
    Regex.match?(~r/[Mm]icrosoft/, osrelease()) and not File.exists?("/.dockerenv")
  end

  defp osrelease() do
    case File.read("/proc/sys/kernel/osrelease") do
      {:ok, text} -> text
      _ -> "unknown"
  end
```

takasehideki: Thanks! This code is smarter than mine!!
Obviously there is no problem at all, but I think we need to make sure this works properly. I will be able to check it early next week. please wait a moment🙏

takasehideki:
I improved my code according to the cool suggestion from @fhunleth 
And also, I've reconfirmed that this improvement works properly in 4 environments mentioned above.
So please check this PR again.

fhunleth (approved):
Looks good and thanks for testing again!

mobileoverlord (approved):

danifv:
Does this PR fix the issue for sure? I just ran into this problem while using WSL2 without docker.  
Removing `not` from https://github.com/nerves-project/nerves/blob/5cb48faa0f1d640113b5193c240b0b2132251dd9/lib/nerves/utils/wsl.ex#L30 appears to fix the issue.

PR #572, opened 2020-12-08
Fix typos

kianmeng:


fhunleth (approved):
Wow, thank you for fixing those!

PR #573, opened 2020-12-10
Devel/error handling

metelik:
A fix for an unhandled error functiona clause

### DIFF
--- a/lib/nerves/utils/http_client.ex
+++ b/lib/nerves/utils/http_client.ex
@@ -69,8 +69,9 @@ defmodule Nerves.Utils.HTTPClient do
     ]
 
     http_opts =
-      [timeout: :infinity, autoredirect: false] ++
-        Nerves.Utils.Proxy.config(url) ++ Keyword.get(opts, :http_opts, [])
+      [timeout: :infinity, autoredirect: false]
+      |> Keyword.merge(Nerves.Utils.Proxy.config(url))
+      |> Keyword.merge(Keyword.get(opts, :http_opts, []))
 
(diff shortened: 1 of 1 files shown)


lib/nerves/utils/http_client.ex
```diff
@@ -69,14 +69,19 @@ defmodule Nerves.Utils.HTTPClient do
...
+      |> Keyword.merge(Keyword.get(opts, :http_opts, []))
 
     opts = [stream: :self, receiver: self(), sync: false]
     :httpc.request(:get, {String.to_charlist(url), headers}, http_opts, opts, :nerves)
     {:noreply, %{s | url: url, caller: from, opts: opts, progress?: progress?}}
   end
 
+  def handle_info({:http, {_ref, {:error, {:failed_connect, [{:to_address, _address}, _reason]}}}}, s) do
+    GenServer.reply(s.caller, {:error, :failed_connect})
+  end
```
jjcarstens, line 83:
```suggestion
  def handle_info({:http, {_ref, {:error, {:failed_connect, _}} = err}}, s) do
    GenServer.reply(s.caller, err)
  end
```
Maybe we should report back more in the error so the user can have a bit more details about the connection failure?

metelik: re-worked as suggested :)

jjcarstens:
Thanks @metelik!

Can you rebase of main and merge your 2 commits? I'm not sure why its showing 0ea6e22 as a new commit here, but that should already be there and not show in this PR

fhunleth:
Rebased and manually merged.

PR #575, opened 2020-12-15
Move vm.args.eex check to `Nerves.Release.init/1`

### LINKED ISSUE #574: vm.args.eex check doesn't work with custom `rel_templates_path` for release
### Environment

* Elixir version (`elixir -v`): 1.11.2
* Nerves environment: (`mix nerves.env --info`)
==> nerves
|nerves_bootstrap| Environment Package List

  No packages found
|nerves_bootstrap| Loadpaths Start

Nerves environment
  MIX_TARGET:   host
  MIX_ENV:      dev

NERVES_SYSTEM is unset
NERVES_TOOLCHAIN is unset
|nerves_bootstrap| Environment Variable List
  target:     host
  toolchain:  unset
  system:     unset
  app:        /Users/benni/Projects/ikea_umbrella

|nerves_bootstrap| Loadpaths End
* Additional information about your host, target hardware or environment that
  may help

### Current behavior

```
==> nerves

Nerves environment
  MIX_TARGET:   rpi3
  MIX_ENV:      dev

** (Mix)   rel/vm.args needs to be moved to rel/vm.args.eex
```

### Expected behavior

It should check `rel_templates_path` and not the hardcoded `rel/vm.args.eex`.

Moving the check into `Nerves.Release.init/1` should resolve this problem.

LostKobrakai:
This should be all needed to allow for alternative `rel_templates_path`s.

Will try this out hopefully later today.

I'm wondering though if the error message could be a bit more generic. Seems like it was mostly added for the distillery -> releases change, but that's not the only reason the file might be missing.

Close #574

### DIFF
--- a/lib/nerves/release.ex
+++ b/lib/nerves/release.ex
@@ -6,6 +6,14 @@ defmodule Nerves.Release do
   def init(%{options: options} = release) do
     opts = Keyword.merge(options, release_opts())
 
+    vm_args = Mix.Release.rel_templates_path(release, "vm.args.eex")
+
+    if !File.exists?(vm_args) do
+      Mix.raise("""
+        rel/vm.args needs to be moved to #{vm_args}
+      """)
+    end
+
     release = %{
       release
       | options: opts,
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -39,16 +39,6 @@ defmodule Mix.Tasks.Firmware do
 
     check_nerves_toolchain_is_set!()
 
-    vm_args =
-      File.cwd!()
-      |> Path.join("rel/vm.args.eex")
-
-    if !File.exists?(vm_args) do
-      Mix.raise("""
-        rel/vm.args needs to be moved to rel/vm.args.eex
-      """)
-    end
-
     # By this point, paths have already been loaded.
     # We just want to ensure any custom systems are compiled
     # via the precompile checks


lib/nerves/release.ex
```diff
@@ -6,6 +6,14 @@ defmodule Nerves.Release do
   def init(%{options: options} = release) do
     opts = Keyword.merge(options, release_opts())
 
+    vm_args = Mix.Release.rel_templates_path(release, "vm.args.eex")
```
mobileoverlord, line 9: We should check the release options to see if this option exists instead of calling directly to `Mix.Release.rel_templates_path` since that is elixir 1.11+

lib/nerves/release.ex
```diff
@@ -6,6 +6,14 @@ defmodule Nerves.Release do
...
+
+    if !File.exists?(vm_args) do
+      Mix.raise("""
+        rel/vm.args needs to be moved to #{vm_args}
+      """)
+    end
+
```
mobileoverlord, line 16: I think this check can go away, this was a carry over from migrating from a distillery project to one that uses elixir releases. The rest of the distillery support was removed.

lib/nerves/release.ex
LostKobrakai: If the check below is no longer needed this line is obsolete anyways.

mobileoverlord:
Finished up in #588

PR #586, opened 2021-01-17
Various doc improvements

axelson:
/data should be used instead of /root

Add a link to info about pushing updates remotely to the getting started
guide

Add notes to FAQ:
- Where to store persistent data
- Applying a firmware update manually (command needs tweaking still)
- Pushing firmware updates remotely

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -6,18 +6,21 @@ If not, please let us know in the #nerves channel on [the Elixir-Lang Slack](htt
 
 ## Where can persistent data be stored?
 
-By default the `/data` partition both read/writeable and is not overwritten when new firmware is pushed to the device.
+By default the `/data` partition both read/writeable and is not overwritten when new firmware is pushed to the device. However, if you are using `mix firmware.burn` then the entire SD card is being overwritten each time so you won't have any persistent data.
 
 ## How can I apply a firmware update manually?
 
+Assuming that you have already put a known good firmware inside "/data/known_good.fw" (perhaps with sftp) then you can run the following commands
+
 ```
-iex> cmd("fwup -i /data/known_good.fw --apply --task complete " <>
+iex> cmd("fwup -i /data/known_good.fw --apply --task upgrade " <>
   "--no-unmount -d #{Nerves.Runtime.KV.get("nerves_fw_devpath")}")
+iex> reboot
 ```
 
 ## How do I push firmware updates remotely?
 
-SSH is a good default for local development and is enabled by default (via `mix nerves.new`) with https://github.com/nerves-project/nerves_firmware_ssh
+SSH is a good default for local development and is enabled by default (via `mix nerves.new`) with https://github.com/nerves-project/nerves_ssh (note: previously https://github.com/nerves-project/nerves_firmware_ssh was enabled by default)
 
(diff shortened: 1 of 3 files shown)


docs/FAQ.md
```diff
@@ -4,6 +4,23 @@ This is a collection of questions that often come up as people are getting start
...
 
+## Where can persistent data be stored?
+
+By default the `/data` partition both read/writeable and is not overwritten when new firmware is pushed to the device.
+
+## How can I apply a firmware update manually?
+
+```
+iex> cmd("fwup -i /data/known_good.fw --apply --task complete " <>
+  "--no-unmount -d #{Nerves.Runtime.KV.get("nerves_fw_devpath")}")
```
axelson, line 15: I built this command by looking at what https://github.com/nerves-project/nerves_firmware_ssh/blob/86e43429c87544c7d18a478bfd200df5667466ff/lib/nerves_firmware_ssh/fwup.ex does. However, this command does not work as-is (testing so far on nerves_system_vultr). On boot I get the following errors:
```
19:49:53.921 [error] FAT-fs (vda1): codepage cp437 not found
19:49:53.921 [warn]  erlinit: Cannot mount /dev/vda1 at /mnt: Invalid argument
19:49:53.921 [error] EXT4-fs (vda4): ext4_check_descriptors: Block bitmap for group 0 not in group (block 4294967295)!
19:49:53.921 [error] EXT4-fs (vda4): group descriptors corrupted!
19:49:53.921 [warn]  erlinit: Cannot mount /dev/vda4 at /root: No error information
19:49:53.924 [info]  mount: mounting /dev/vda4 on /root failed: No error information
19:49:53.925 [warn]  heart: kernel watchdog activated (interval 5s)
19:49:53.925 [error] EXT4-fs (vda4): ext4_check_descriptors: Block bitmap for group 0 not in group (block 4294967295)!
19:49:53.925 [error] EXT4-fs (vda4): group descriptors corrupted!
19:49:53.925 [warn]  Ignoring non-zero exit status from mount ["-t", "ext4", "-o", "rw", "/dev/vda4", "/root"]
```

Also, I believe we'll need to tell the user to run `reboot/0`.

fhunleth: If you change the task to "upgrade", it should work.

axelson: Ah, that does work :+1:. Didn't expect "upgrade" to be a task based on the `fwup -l` output:
```
fwup -l -i good.fw
complete
upgrade.a
upgrade.b
upgrade.unexpected
upgrade.wrongplatform
```
I thought you would need to do `upgrade.a` or b

axelson:
Okay, made some slight updates to this, and changed the task to "upgrade"

docs/FAQ.md
fhunleth: Yeah, the task passed on the commandline is a prefix match for the first applicable task in the fwup.conf file. That lets you force fwup to upgrade the A or B partition or let fwup decide which is the right one.

fhunleth:
Looks good! Thanks.

docs/FAQ.md
axelson: Okay, that makes sense :+1:

PR #587, opened 2021-01-19
Remove strange hack in http client

ConnorRigby:
I tested this on all of our latest systems from the last few versions. I'm not sure what this was trying to accomplish, but i don't think it's necessary anymore and I can't sleep at night knowing it exists.

### DIFF
--- a/lib/nerves/utils/http_client.ex
+++ b/lib/nerves/utils/http_client.ex
@@ -27,7 +27,6 @@ defmodule Nerves.Utils.HTTPClient do
       uri
       |> URI.to_string()
       |> URI.encode()
-      |> String.replace("+", "%2B")
 
     get(pid, url, opts)
   end


fhunleth (approved):

mobileoverlord:
The `+` symbol is part of the resetved set that according to the spec should be percent encoded. https://tools.ietf.org/html/rfc3986

I've tested this in iex and URI.encode alone does not encode them. The `+` specifically has special meaning to which it will be interpreted as a space. If anything the code should possibly be updated to handle more of the reserved characterset.

fhunleth:
@mobileoverlord The spec you referenced has quite a few reserved characters. Isn't it odd to single out `+` for handling and not any of the other ones?

mobileoverlord:
I think this was an accidental addition for an isolated case that wasn't tied back until now. As I mentioned, if anything we should probably encode the others in the reserved set instead of removing just this one.

fhunleth:
@ConnorRigby Do you want to rework the PR to include the other characters?

Just as a side comment, is something broken when it comes to URI encoding? Like why do we have to deal with this. It seems like this should apply to everyone?

fhunleth (changes requested):
Changed approval based on comments.

fhunleth:
Closing out of lack of interest in cleaning it up.

PR #589, opened 2021-01-22
CI: bump Erlang/Elixir, add checks

fhunleth:


.circleci/config.yml
```diff
@@ -1,5 +1,5 @@
 otp_23_image: &otp_23_image
-  image: erlang:23.1
+  image: erlang:23.2.1
```
jjcarstens, line 2: I am guessing 23.2.3 isn't available here?

fhunleth: No. This is the whole reason I've been switching to the hex docker images for other projects. I just didn't have the time this morning rework the CI script. I main goal was to update ex_doc deps since I was pushing doc updates and then I merged the other little improvements.

jjcarstens (approved):

PR #605, opened 2021-03-27
Bug fix for double encoding of URL query parameters in OTP-23.2+

### LINKED ISSUE #604: Double percent encoding of URLs leads to OTP version dependent results
When downloading nerves build artifacts using the URI resolver, the query string is double encoded. This works on OTP-23.1.5 and below but fails on OTP-23.2 and above because starting in OTP-23.2, httpc no longer removes the double encoding. This translates to Nerves requesting invalid URLs when fetching build artifacts.

### Environment
* Elixir versions impacted: tested on 1.10.4, but should impact any version built with an impacted OTP version
* Erlang versions impacted: OTP-23.2 and above
* Nerves environment:
  * Nerves: 1.6.6
  * Nerves Bootstrap: 1.10.2

Nerves 1.7.4 (latest) should be impacted as well, none of the functions in question have changed.

### Current behavior

Nerves requests an invalid URI due to the double percent encoding of query parameters.

### Expected behavior

Nerves should only percent encode a URI once.

### Context
We store our proprietary nerves system builds in AWS S3.
The proprietary nerves system repo is configured to provide query parameters through the nerves_package mix project option. E.g.
```elixir
def project do
  [
    nerves_package: [
      ...
      artifact_sites: [
        {:prefix, prefix, query_params: %{"X-Amz-Credential" => "a/b/c/d", ...}}
        ...
      ]
    ]
  ]
end
```

AWS requires a parameter that has `/` as part of the query string value (example shown above).
When fetching the pre-built artifact, [Nerves encodes](https://github.com/nerves-project/nerves/blob/main/lib/nerves/artifact/resolvers/uri.ex#L21) the `:query_params` option using `URI.encode_query` which results in a uri like this:
```elixir
uri = %URI{
  authority: "releases.s3.amazonaws.com",
  fragment: nil,
  host: "s3.amazonaws.com",
  path: "/nerves_system.tar.gz",
  port: 443,
  query: "X-Amz-Credential=a%2Fb%2Fc%2Fd",
  scheme: "https",
  userinfo: nil
}
```
At this point you can see the query has been percent-encoded once (`"/"=> %2F`).
Nerves then prepares the URI for fetching by [encoding the entire URI string](https://github.com/nerves-project/nerves/blob/main/lib/nerves/utils/http_client.ex#L28), resulting in a double encoded query string.
```elixir
uri
|> URI.to_string()
|> URI.encode()
|> String.replace("+", "%2B")
"https://s3.amazonaws.com/nerves_system.tar.gz?X-Amz-Credential=a%252Fb%252Fc%252Fd"
```

Once this URL string is [passed to httpc](https://github.com/nerves-project/nerves/blob/main/lib/nerves/utils/http_client.ex#L71) as a request, httpc calls `:uri_string.normalize` which is where the different behavior comes in.
Starting with OTP-23.2, `:uri_string.normalize` will no longer decode the `%25` values from the string, causing httpc to pass an invalid URL to AWS. This was confirmed in OTP-23.2 and OTP-23.3. Prior to OTP-23.2, `:uri_string.normalize` would decode these values (confirmed in OTP-23.0, OTP-23.1, and OTP-23.1.5).

The OTP-23.2 additions make the query parser and normalization procedure much more sophisticated than previous versions. It works by first creating the fragments and then decoding as needed, with different reserved character sets for each fragment. Similarly, when composing a URI, it performs fragment specific encoding like the following
```elixir
%{scheme: "http", host: "a b", path: "", query: :uri_string.compose_query([{"a", "a/b/c"}])} |> :uri_string.recompose
# See how the query string is not double encoded
"http://a%20b?a=a%2Fb%2Fc"
```

### Recommendation
* Drop the `URI.encode` of the entire URL and the replacement of `+` with `%20` in the [http client](https://github.com/nerves-project/nerves/blob/main/lib/nerves/utils/http_client.ex#L28)
* Change the contract for `HTTPClient.get` to require a valid URL
* Update [the uri resolver](https://github.com/nerves-project/nerves/blob/main/lib/nerves/artifact/resolvers/uri.ex#L20) to encode the URI before adding query params to it.
* Update [github API resolver](https://github.com/nerves-project/nerves/blob/main/lib/nerves/artifact/resolvers/github_api.ex#L40) to  encode the URL before passing it to the client

If this sounds good, I can open a PR with these changes. I want to confirm that I didn't miss any addition context.

jthacker:
Resolves #604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+* Bug fixes
+  * Fixes an issue where query parameters would be percent-encoded twice.
+    Packages that use `query_params` argument option to `artifact_sites` could
+    be impacted. For example, packages storing build artifacts in AWS S3
+    require the `X-Amz-Credential` query parameter key whose value
+    includes the reserved character `/`. This symbol is double encoded to
+    `%252F`. This failed on systems with Erlang OTP-23.2 and above.
+    See https://github.com/nerves-project/nerves/issues/604 for additional context.
+
 ## v1.7.4
 
 * Experimental features
--- a/lib/nerves/artifact/resolvers/github_api.ex
+++ b/lib/nerves/artifact/resolvers/github_api.ex
@@ -41,6 +41,8 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
 
     Nerves.Utils.Shell.info("  Downloading artifacts from #{url}")
 
+    url = URI.encode(url)
+
     result =
       with {:ok, data} <- HTTPClient.get(http_pid, url, headers: [auth_header], progress?: false),
            %{"assets" => assets} <- Utils.json_decode(data),
--- a/lib/nerves/utils/http_client.ex
+++ b/lib/nerves/utils/http_client.ex
@@ -22,12 +22,7 @@ defmodule Nerves.Utils.HTTPClient do
   end
 
   def get(pid, %URI{} = uri, opts) do
-    url =
-      uri
-      |> URI.to_string()
-      |> URI.encode()
-      |> String.replace("+", "%2B")
-
+    url = URI.to_string(uri)
     get(pid, url, opts)
   end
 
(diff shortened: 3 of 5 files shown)


fhunleth:
Thanks! The PR looks good and I tested it against the public and private artifact locations that I'm using.

fhunleth (approved):

PR #619, opened 2021-06-01
Add new mix task for generating a gdb wrapper script

ConnorRigby:
This is required for programs like `gdb`, `debuginfod` and similar to automatacally locate debug symbols when present

### DIFF
--- a/lib/mix/tasks/firmware.gen.gdb.ex
+++ b/lib/mix/tasks/firmware.gen.gdb.ex
@@ -1,0 +1,37 @@
+defmodule Mix.Tasks.Firmware.Gen.Gdb do
+  use Mix.Task
+  import Mix.Nerves.Utils
+  alias Mix.Nerves.Preflight
+
+  @script_name "gdb.sh"
+
+  @shortdoc "Generates a helper shell script for using gdb to analyze core dumps"
+
+  @moduledoc """
+  Generates a helper shell script for using gdb to analyze core dumps
+  
+  This script may be used on its own or used as a base for more complicated debugging.
+  It saves the script to #{@script_name}.
+  """
+  @spec run(keyword()) :: :ok
+  def run(_args) do
+    Preflight.check!()
+    system_path = check_nerves_system_is_set!()
+    check_nerves_toolchain_is_set!()
+
+    gdb_script_contents =
+      Application.app_dir(:nerves, "priv/templates/script.run-gdb.sh.eex")
+      |> EEx.eval_file(assigns: [nerves_system: system_path])
+
+    if File.exists?(@script_name) do
+      Mix.shell().yes?("OK to overwrite #{@script_name}?") || Mix.raise("Aborted")
+    end
+
+    Mix.shell().info("""
+    Writing #{@script_name}...
+    """)
+
+    File.write!(@script_name, gdb_script_contents)
+    File.chmod!(@script_name, 0o755)
+  end
+end
--- a/mix.exs
+++ b/mix.exs
@@ -29,7 +29,7 @@ defmodule Nerves.MixProject do
   end
 
   def application do
-    [extra_applications: [:ssl, :inets]]
+    [extra_applications: [:ssl, :inets, :eex]]
   end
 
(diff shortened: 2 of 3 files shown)


fhunleth:
I think we might be able to do this by updating the systems to contain a debug directory filled with build_id -> executable/library symlinks. There shouldn't be any need to extract symbols from the binaries since gdb can look at the binaries and ignore the data/code/etc. sections.

docs/Debugging.md
```diff
@@ -0,0 +1,30 @@
+# Debugging Nerves Systems
+
+## Obtaining a core dump from vm crashes
+
+To obtain a core dump if beam crashes, you'll need a few things first:
+
+* core dumps enabled in the Linux Kernal (this is on by default for all officially supported targets)
```
fhunleth, line 7:
```suggestion
* core dumps enabled in the Linux Kernel (this is on by default for all officially supported targets)
```

docs/Debugging.md
```diff
@@ -0,0 +1,30 @@
...
+  * `CONFIG_ELF_CORE` and `CONFIG_COREDUMP` both must be `=Y` in the kernel config
+* process core size is greater than zero
+  * this can be checked with `cmd("ulimit -c")` and set with `cmd("ulimit c unlimited")`
```
fhunleth, line 10:
```suggestion
  * this can be checked with `cmd("ulimit -c")` and set with `cmd("ulimit -c unlimited")`
```

priv/templates/script.gdb.sh.eex
```diff
@@ -0,0 +1,39 @@
...
+[ -n "$ERTS_INCLUDE_DIR" ] || (echo "Error: missing environment variable $ERTS_INCLUDE_DIR"; help)
+
+[ -n "$CROSSCOMPILE" ] || (echo "Warning: missing environment variable $CROSSCOMPILE")
+GDB=$CROSSCOMPILE-gdb
+[ -f "$GDB" ] || (echo "Error: gdb is not available in this toolchain"; help)
+
+$GDB --core="$CORE" \
+     --nx \
+     --init-eval-command="set sysroot $NERVES_SDK_SYSROOT" \
+     "$ERTS_INCLUDE_DIR/../bin/beam.smp"
```
fhunleth, line 39: Wasn't Build-ID supposed to make it so that we didn't need to specify the path?

priv/templates/script.gdb.sh.eex
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+#
+# Upload new firmware to a target running nerves_firmware_ssh
+#
+# Usage:
+#   gdb.sh [path to executable] [path to core dump]
```
fhunleth, line 7:
```suggestion
#   gdb.sh <path to core dump> [path to executable]
```

lib/mix/tasks/firmware.gen.gdb.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Mix.Tasks.Firmware.Gen.Gdb do
+  use Mix.Task
+  import Mix.Nerves.Utils
+  alias Mix.Nerves.Preflight
+
+  @script_name "gdb.sh"
+
+  @shortdoc "Generates a shell script for starting a gdb session to analyse a crash"
```
fhunleth, line 8:
```suggestion
  @shortdoc "Generates a helper shell script for using gdb to analyze core dumps"
```

lib/mix/tasks/firmware.gen.gdb.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  @moduledoc """
+  Creates a shell script for starting a gdb session to analyse a crash.
+  This script may be used on its own or used as a base for more complicated debugging
+  It saves the script to #{@script_name}.
```
fhunleth, line 13:
```suggestion
  Generates a helper shell script for using gdb to analyze core dumps
  
  This script may be used on its own or used as a base for more complicated debugging.
  It saves the script to #{@script_name}.
```

PR #631, opened 2021-06-26
Update installation doc Elixir/OTP versions

axelson:
The current nerves versions and systems require OTP 24 to avoid the host and target have different OTP versions error, so the installation guide should be updated as well.

Reported on slack


kylepotts (approved):

fhunleth:
Thanks for fixing this!!!

axelson:
You're welcome!

PR #638, opened 2021-07-21
Allow specifying firmware path with firmware.unpack

jjcarstens:
`mix firmware.unpack` would only look for a firmware that was currently built
for a project. Sometimes you may have multiple firmware files to inspect that
were built elsewhere and this would be a convenient tool to inspect them.

This adds `--fw path/to/.fw` support for those cases. It also changes the default
outpath to be the firmware file name with `.unpacked` appended. (i.e. `hello_nerves.fw -> hello_nerves.unpacked`)

### DIFF
--- a/lib/mix/tasks/firmware.unpack.ex
+++ b/lib/mix/tasks/firmware.unpack.ex
@@ -10,10 +10,15 @@ defmodule Mix.Tasks.Firmware.Unpack do
 
   ## Usage
 
-      mix firmware.unpack [output directory]
+      mix firmware.unpack [--output output directory] [--fw path to firmware]
 
-  If not supplied, the output directory will be based off the OTP application
-  name.
+  ## Command line options
+
+    * `--fw` - (Optional) The path to the .fw file for unpacking.
+      Defaults to `Nerves.Env.firmware_path/1`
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/firmware.unpack.ex
```diff
@@ -24,54 +29,37 @@ defmodule Mix.Tasks.Firmware.Unpack do
   # Unpack it
   mix firmware.unpack firmware_contents
 
+  # Unpack specified fw file
+  mix firmware.unpack --fw hello_nerves.fw
+
   # Inspect it
   ls firmware_contents
```
fhunleth, line 28: This line isn't correct any more.

fhunleth (commented):
It feels really odd to me having the input being an option and the output directory being a positional arg now. Since changing the default output name is an API breaking change anyway, how about making the output directory an option too?

jjcarstens:
@fhunleth Changed to support `--output` option as well

fhunleth (approved):
Looks good to me! I like you changed the default output directory to something that makes more sense.

mobileoverlord (approved):

PR #643, opened 2021-07-23
Don't set xattrs when unpacking archives

fhunleth:
On some systems (Arch Linux) xattrs needs additional permissions. Since
`mix firmware.unpack` is intended as a debug tool, having accurate
xattrs hardly seems important so this disables them.


jjcarstens (approved):

fhunleth:
@ConnorRigby - Since you're our resident Arch Linux person, did you ever work around xattrs? This isn't an issue for me.

ConnorRigby (approved):
i just double checked - `mix firmware.unpack` works fine for me without this. 
Tested the change out and it seems to still work, so lgtm
:sunglasses:

PR #645, opened 2021-07-25
Add docs for configuring kernel parameters

amclain:
Related to nerves-project/nerves_system_br#570

This PR adds documentation for how to use `sysctl` to configure kernel parameters, which was introduced in nerves-project/nerves_system_br#570.


fhunleth:
Thanks! 

I added some information that you need a supporting `nerves_runtime` to use this and that Nerves systems may or may not provide a `/etc/sysctl.conf`. 

It now feels caveat'd enough that I went ahead and merged. I haven't pushed updated hex docs yet.

PR #648, opened 2021-08-04
Update Ubuntu installation instructions.

YurkoHoshko:
Updated Ubuntu installation instruction for newer Ubuntu versions.

As per https://askubuntu.com/questions/1241217/package-libwxgtk3-0-dev-has-no-installation-candidate-on-ubuntu-20-04 - the package was renamed from libwxgtk3.0-dev to libwxgtk3.0-gtk3-dev.

Tested on Windows 10 + WSL2 with Ubuntu 20.04


fhunleth:
I also confirmed the new package name. Thanks for sending the PR!

PR #652, opened 2021-08-31
Add firmware build success message

fhunleth:
This adds a short message for what to do next after running `mix
firmware`. The goal is to have a simple reminder for what to do next.

It looks like this:

```sh
...
Building /Users/fhunleth/git/nerves-project/nerves_examples/blinky/_build/rpi0_dev/nerves/images/blinky.fw...
Firmware built successfully! 🎉

Now you may install the firmware on a MicroSD card using `mix burn` or upload
the firmware to a device with `mix upload` or `mix firmware.gen.script`/`./upload.sh`.
```

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -55,13 +55,20 @@ defmodule Mix.Tasks.Firmware do
   end
 
   @doc false
-  def result({_, 0}), do: nil
+  def result({_, 0}) do
+    Mix.shell().info("""
+    Firmware built successfully! 🎉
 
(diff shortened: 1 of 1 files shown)


fhunleth:
@jjcarstens I just realized how abrupt our builds end so I figured that I'd add this. What do you think? Good, bad, shorter?

jjcarstens:
I think its probably fine. If anything, I think adding the firmware UUID would be nice. And its probably worth noting that they will get this message even if doing `mix firmware.burn`, but thats prob inconsequential

jjcarstens (approved):

PR #655, opened 2021-09-05
Add docs for using the CLI

fhunleth:
In addition to just having more documentation, the idea is to replace the RingLogger info in the MOTD with a URL to this page.

### DIFF
--- a/docs/Using the CLI.md
+++ b/docs/Using the CLI.md
@@ -1,0 +1,366 @@
+# Using the CLI
+
+Nerves greets you with Elixir's IEx prompt. This prompt is your main entry point
+to interacting with Elixir, your program and hardware. This chapter focuses on
+Nerves-specific use of the IEx prompt.
+
+## Viewing log messages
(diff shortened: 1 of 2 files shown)


docs/Using the CLI.md
```diff
@@ -0,0 +1,365 @@
...
+
+```elixir
+config :nerves, :erlinit,
+  ctty: "null",
+  alternate_exec: "/usr/bin/run_erl /tmp/ /tmp exec"
+```
+
+The `:alternate_exec` key is optional here. It calls
+[`run_erl`](https://erlang.org/doc/man/run_erl.html) to log console output to a
+file in `/tmp`. This is useful if code calls `IO.puts` rather then `Logger`.
```
jjcarstens, line 302:
```suggestion
file in `/tmp`. This is useful if code calls `IO.puts` rather than `Logger`.
```

jjcarstens (approved):

PR #661, opened 2021-10-10
Update phoenix config in the User Interfaces doc

mnishiguchi:
Updates for Phoenix 1.6 based on what I learned at https://github.com/nerves-project/nerves_examples/pull/220 and my [memo](https://dev.to/mnishiguchi/elixir-poncho-project-with-nerves-firmware-and-phoenix-liveview-ui-48k) on it.

## Changes

- Focus on a poncho project structure so that we can simplify the explanation
- Update the instructions for Phoenix 1.6
- List all the necessary UI configuration in the firmware project's `config/target.exs` instead of importing them the UI project and override (simplicity; consistency with `hello_nerves` example)
- Misc clean up

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -1,29 +1,48 @@
 # User Interfaces
 
-## Phoenix web interfaces
+## Phoenix web interface
 
-The Phoenix web framework makes an excellent companion to Nerves-based devices
+The [Phoenix] web framework makes an excellent companion to [Nerves]-based devices
 that need to serve content over HTTP directly from the device. For example, a
 device with no display might provide administration and configuration
 interfaces intended to be accessed from a computer or mobile device.
 
+[Nerves]: https://www.nerves-project.org/
+[Phoenix]: http://www.phoenixframework.org/
+
 ### Choosing a project structure
 
-Although Nerves supports umbrella projects, the preferred project structure is
-to simply use separate Mix projects, side-by-side with path dependencies
-between them, in the same source code repository. We call this a "poncho
-project" structure. For the reasoning behind this, please see the original
-[blog post describing poncho projects].
+There are two commonly used project structures for Nerves-based devices that uses
+Phoenix web interface:
+
+- [poncho project structure]
+- [umbrella project structure]
+
+Although Nerves supports both, the preferred project structure is what we call
+poncho project structure. For the reasoning behind this, please see the original
+[blog post describing poncho projects]. The following steps assume that we use
+the poncho project structure.
+
+Using the poncho project structure, we simply use separate [Mix projects],
+side-by-side with path dependencies between them, in the same source code repository.
 
 [blog post describing poncho projects]: https://embedded-elixir.com/post/2017-05-19-poncho-projects/
+[poncho project structure]: http://embedded-elixir.com/post/2017-05-19-poncho-projects/
+[umbrella project structure]: https://elixir-lang.org/getting-started/mix-otp/dependencies-and-umbrella-projects.html
+[Mix projects]: https://hexdocs.pm/mix/Mix.html
 
-### Using a poncho project structure
+### Create a poncho project
 
 First, we generate the two new Mix projects in a containing directory:
 
 ```bash
+# Create a container directory called "my_app"
 mkdir my_app && cd my_app
+
+# Create a Nerves firmware project called "my_app_firmware"
 mix nerves.new my_app_firmware
+
+# Create a Phoenix 1.6 UI project called "my_app_ui"
 mix phx.new my_app_ui --no-ecto --no-mailer
 ```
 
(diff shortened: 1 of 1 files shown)


nbw:
@mnishiguchi yes, `mix assets.deploy`. You're right. apologies.

nbw:
My code seems to building 🤞  for now. What errors are you seeing?

I've pushed my code to github (removed my wifi creds, so you'd have to fill those in to get it working). I wasn't planning on pushing it yet so excuse the lack of comments or description: 
https://github.com/nbw/nerves_weather

Comparing at your target.exs file for your firmware: [your target.exs](https://github.com/mnishiguchi/nerves_examples/blob/mnishiguchi/phoenix-1-6/hello_phoenix/firmware/config/target.exs) to mine [my config.exs](https://github.com/nbw/nerves_weather/blob/main/nerves_weather_firmware/config/config.exs) you can see the difference. The point of interest in this case being: `watchers: []`. 

It might not be important, but I also used `http: [ip: {0, 0, 0, 0}, port: 80],` since the dev.exs config generated by phoenix said 

> Change to `ip: {0, 0, 0, 0}` to allow access from other machines. 

Not sure if it made a difference in the end. 

---

Originally, when created a phoenix 1.6 project for the first time (without touching the phoenix config at all) I confirmed what the error was by:

1. ssh in to my nerves machine (`ssh nerves.local` if you're setup)
2. type `log_attach` from terminal to see logs
3. there would be an error related to `Esbuild` and `:install_and_run`

mnishiguchi:
Thanks for your code @nbw !
Looks like one major difference is that

- mine places all the UI config for the firmware in the firmware config
- yours imports the UI config in the firmware config and overrides some

Screen Shot 2021-10-12 at 8 22 08 AM

I chose to explicitly list all the necessary settings in the firmware config because I thought it would be easier to reason about how the UI is configured in the long run.

As of now, I still get this `esbuild` error on my firmware.

Screen Shot 2021-10-12 at 8 13 23 AM

Maybe it is easier to just not load `esbuild` on the target machine so we do not need to worry about `esbuild` in the firmware runtime. What do you think?

```elixir
{:esbuild, "~> 0.2", runtime: Mix.env() == :dev && Mix.target() == :host},
```

mnishiguchi:
Yeah for now I'll go with  `{:esbuild, "~> 0.2", runtime: Mix.env() == :dev && Mix.target() == :host}`, then we can keep on updating it 😃 

Btw you can still compile assets and do normal phoenix development with `{:esbuild, "~> 0.2", runtime: Mix.env() == :dev && Mix.target() == :host}`. It only makes sure `esbuild` is not loaded at runtime on Raspberry Pi (target).

nbw:
了解です。英語ネイティブなんで、順番のみならず文法も確認します。I will test when I have a chance (tomorrow probably), but I did some double checking and I have to admit that my config does _not_ work. 

My nerves machine is working fine because I'm compiling the assets beforehand (then build firmware, then burn). However, I ssh'ed into the machine and checked the logs just in case and I'm still getting:

```
[error] Task #PID<0.1967.0> started from NervesWeatherUiWeb.Endpoint terminating
** (UndefinedFunctionError) function Esbuild.install_and_run/2 is undefined (module Esbuild is not available)
    Esbuild.install_and_run(:default, ["--sourcemap=inline", "--watch"])
    (phoenix 1.6.2) lib/phoenix/endpoint/watcher.ex:19: Phoenix.Endpoint.Watcher.watch/2
    (elixir 1.12.2) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (stdlib 3.15.2) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```

So my original idea to set:

```
 watchers: []
```

does not work. The nerves machine continues to run because it has the assets it needs, but behind the scenes its spewing EsBuild errors. 

So in my Phoenix project's dev.exs I commented out the Esbuild line entirely.

```
  watchers: [
    # esbuild: {Esbuild, :install_and_run, [:default, ~w(--sourcemap=inline --watch)]}
  ]
```

Unfortunately this means on when your developing locally you either have to uncomment that line or recompile assets manually using `mix assets.deploy`

fhunleth:
@mnishiguchi and @nbw - How's this coming? It looks like it's close and I'd love to get this posted so that people can start transitioning to Phoenix 1.6.

fhunleth:
Sounds good. Yes, let's remove the umbrella app instructions and get this one working. It can be added back in a future PR.

mnishiguchi:
@fhunleth I tested the instructions I put in the docs more than a few times, by building an app from scratch. I am confident in it. If you are good, I'd say the doc is good to go.

Then we can continue to fine-tune it based on upcoming reports, complaints etc.

jjcarstens:
I really don't know much about the new ESbuild process. It seems like that is required in Phoenix now? (Maybe it can be skipped at runtime?)

If not, you'd need to set the application env to point to a different path

https://github.com/phoenixframework/esbuild/blob/e2d01c87ecd60d02ab8151dcbd51a24a13495f21/lib/esbuild.ex#L109

And then prob confirm this can evaluate to the appropriate binary when on device:

https://github.com/phoenixframework/esbuild/blob/e2d01c87ecd60d02ab8151dcbd51a24a13495f21/lib/esbuild.ex#L220-L239

But...I feel like that shouldn't be needed at runtime. If it is, we might need to adjust Phoenix

fhunleth:
@mnishiguchi and @nbw Thanks for figuring this out and for the doc updates!

PR #667, opened 2021-11-14
Add an additional link to the targets page

axelson:
This link is a bit easier to scan than the existing link that is within the text.

## Before:
Screen Shot 2021-11-14 11-33-10

## After:
Screen Shot 2021-11-14 11-34-22


fhunleth:
Looks good!

PR #674, opened 2021-12-15
Verify artifact downloads

jjcarstens:
With OTP 24, it now warns you when attempting an HTTPS request without any verification, although TLS still allows it.
    
This takes the hint to get a little more secure when downloading Nerves artifacts and attempts to establish authenticity from trusted certs in `:castore`, which should include the most common github release downloads

### DIFF
--- a/lib/nerves/utils/http_client.ex
+++ b/lib/nerves/utils/http_client.ex
@@ -59,7 +59,17 @@ defmodule Nerves.Utils.HTTPClient do
     ]
 
     http_opts =
-      [timeout: :infinity, autoredirect: false]
+      [
+        timeout: :infinity,
+        autoredirect: false,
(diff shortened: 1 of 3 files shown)


fhunleth:
This is needed, but the logic needs to at least be switched. If the archive is coming from an official location, then we need to verify that it is, indeed, coming from that location by checking the certs.

jjcarstens:
@fhunleth Updated.

One thing I am unsure of is if we need to exclude the `ssl` options if the URL is `http`

fhunleth:
I don't know what happens with `http` either, but it's 2021 and I don't feel like we need to spend time verifying that it works.

fhunleth (approved):
It works for me. Thanks!

Issue #679, opened 2021-12-15
Adjust burn task to detect if drive already has Nerves image


jjcarstens:
By default `mix burn` (and subsequently `mix firmware.burn`) will completely overwrite the SD card because it defaults to using the `complete` task, losing all data on it

However, it may be possible to detect the SD card already has an image on it and adjust the task to use `upgrade` instead which would preserve the `/data` partition on disk like when upgrading OTA via SSH or NervesHub

axelson:
Oooh, that would be nice! Another option would be for `mix firmware.burn` to not allow overwriting an existing nerves image and instead require `mix firmware.burn --overwrite` to actually overwrite the image.

jjcarstens:
That's precisely what I was thinking @axelson! Default to `upgrade`, which would fail if there is no Uboot env (according to Frank), then fallback to complete. And also allow `--overwrite` option to force it

PR #682, opened 2021-12-28
Change `burn` task to Upgrade by default

jjcarstens:
This changes the `mix burn` (and subsequently `mix firmware.burn`) to use the FWUP `upgrade` task by default. If the upgrade fails because the device does not actually have upgradable firmware on it, then the task will fallback to using the `complete` task to overwrite the device.

You can also now specify `--overwrite` to force the `complete` task and wipe the data on the device with the new firmware.

see #679

### DIFF
--- a/lib/mix/tasks/burn.ex
+++ b/lib/mix/tasks/burn.ex
@@ -1,10 +1,10 @@
 defmodule Mix.Tasks.Burn do
   use Mix.Task
   import Mix.Nerves.Utils
-  alias Mix.Nerves.Preflight
+  alias Mix.Nerves.{FwupStream, Preflight}
   alias Nerves.Utils.WSL
 
-  @switches [device: :string, task: :string, firmware: :string]
+  @switches [device: :string, task: :string, firmware: :string, overwrite: :boolean]
   @aliases [d: :device, t: :task, i: :firmware]
 
   @shortdoc "Write a firmware image to an SDCard"
@@ -13,8 +13,9 @@ defmodule Mix.Tasks.Burn do
   Writes the generated firmware image to an attached SDCard or file.
 
   By default, this task detects attached SDCards and then invokes `fwup`
-  to overwrite the contents of the selected SDCard with the new image.
-  Data on the SDCard will be lost, so be careful.
+  to upgrade the contents of the selected SDCard with the new image.
+  If the upgrade to the next parition fails, it will then attempt to
+  completely overwrite the SDCard with the new image.
 
   ## Command line options
 
@@ -29,9 +30,13 @@ defmodule Mix.Tasks.Burn do
       convention, the `complete` task writes everything to the SDCard including
       bootloader and application data partitions. The `upgrade` task only
       modifies the parts of the SDCard required to run the new software.
+      Defaults to `upgrade`
 
     * `--firmware <name>` - (Optional) The path to the fw file to use.
       Defaults to `<image_path>/<otp_app>.fw`
+
+    * `--overwrite` - (Optional) Overwrite the contents of the SDCard by
+      forcing the `complete` task. Defaults to `false`
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


fhunleth:
This is really cool. I think this will help so many people. I'll give it a try later today.

lib/mix/tasks/burn.ex
```diff
@@ -71,56 +76,39 @@ defmodule Mix.Tasks.Burn do
...
+    {cmd, args} = cmd_and_args_for_os(os, args, dev)
+
+    shell(cmd, args, stream: FwupStream.new())
+    |> format_result(task)
+    |> case do
+      :failed_not_upgradable ->
+        Mix.shell().info("""
+        #{IO.ANSI.yellow()}
+        FWUP upgrade was attempted but #{dev} does not have a previous firmware.
+        Switching to use the #{IO.ANSI.cyan()}complete#{IO.ANSI.yellow()} task to format the entire card.
```
fhunleth, line 103:
```suggestion
        Device #{dev} either doesn't have firmware on it or has incompatible firmware.
        Going to burn the whole MicroSD card so that it's in a factory-default state.
```

lib/mix/tasks/burn.ex
```diff
@@ -157,4 +145,58 @@ defmodule Mix.Tasks.Burn do
...
+  defp cmd_and_args_for_os(:wsl, args, _dev) do
+    WSL.admin_powershell_command("fwup", Enum.join(args, " "))
+  end
+
+  defp cmd_and_args_for_os(_os, args, _dev), do: {"fwup", args}
+
+  defp format_result({_, 0}, _task), do: :ok
+
+  defp format_result({%FwupStream{output: o}, _}, "upgrade") do
+    if o =~ ~r/fwup: Expecting platform=#{mix_target()} and/ do
```
fhunleth, line 194: This messages comes from whatever the user puts in their `fwup.config`, so it seems fragile. Does this work with a FAT-formatted MicroSD card? Like one that you get fresh from a store?

jjcarstens: Hmm, didn't think about that. I will try with a fresh card shortly

fhunleth (commented):
The switch to the `complete` task isn't working on my MBP:

```
Use 7.4 GiB memory card found at /dev/rdisk5? [Yn]
  0% [                                    ]
fwup: Expecting platform=rpi0 and architecture=arm

FWUP upgrade was attempted but /dev/rdisk5 does not have a previous firmware.
Switching to use the complete task to format the entire card.


  0% [                                    ]
fwup: Cannot open '/dev/rdisk5' for output.
Check file permissions or the read-only tab if this is an SD Card.
```

I think there's a race condition and there needs to be a pause?

`mix burn --overwrite` works, though.

lib/mix/tasks/burn.ex
jjcarstens: I am getting the same message on a newly FAT formatted microSD and USB drive.

I'm wondering if they change here is to not be automatic and just prompt again asking for a confirmation to run the `complete` task? Or...we just always run complete if the `upgrade` task fails?

jjcarstens:
> I think there's a race condition and there needs to be a pause?

Did you MBP prompt you a second time for  a password? I've been unable to reproduce on 3 machines (Linux, MBP, and Mac mini M1), but I get prompted for a password a second time when it does the next task. I'm more just curious now and can easily add in a quick pause to be safe.

lib/mix/tasks/burn.ex
jjcarstens: Technically, even the task can be defined by the user as well and there may not even be an `upgrade` or `complete` task. Maybe this is considered a pure convenience and the only support for this "auto" work is with what's defined in official systems which all have this message and these tasks?

fhunleth:
Before you add a pause, there might be another way. For example, the logic could be put in the `fwup.config` or `fwup` could be extended to let you specify a "fallback" task if the main one errors out. That would prevent the need to authenticate twice. I'm not sure if authenticating twice is bad, though, so long as it works the second time.

I should show you live what I ran into when we pair on the `ssh` issue in case I'm doing something silly.

PR #687, opened 2022-01-17
Add minimum nerves_system_br version for buid env

fhunleth:
- Remove erl_interface references since removed in OTP 23
- Add minimum nerves_system_br version for buid env


jjcarstens:
If we're including nerves_system_br versions, maybe we shouldn't remove the erl_interface reference instead put the version range it was included?

fhunleth:
I looked at that. It has been deprecated for so long that I had no interest in maintaining any use of it.

jjcarstens (approved):

PR #688, opened 2022-01-20
unset erlang vars when building system

jjcarstens:


lib/mix/nerves/shell.ex
```diff
@@ -25,7 +25,16 @@ defmodule Mix.Nerves.Shell do
...
-        {:env, [{'PATH', Mix.Nerves.Utils.sanitize_path() |> to_charlist()}]}
+        {:env,
+         [
+           {'PATH', Mix.Nerves.Utils.sanitize_path() |> to_charlist()},
+           # Unset these Env vars which are set by the host Erlang
+           # and might interfere with the build
+           {'BINDIR', ''},
+           {'MIX_HOME', ''},
+           {'PROGNAME', ''},
+           {'ROOTDIR', ''}
```
fhunleth, line 36:
```suggestion
           {'BINDIR', nil},
           {'MIX_HOME', nil},
           {'PROGNAME', nil},
           {'ROOTDIR', nil}
```

I think it will be safer to unset them rather than set them to the empty string.

jjcarstens: I tried this first but `nil` values are not allowed

https://www.erlang.org/doc/man/os.html#type-env_var_value

But, it does look like we can set it to `false` (though I don't know what that will do)

fhunleth: Ahh. `false` is right: https://www.erlang.org/doc/man/erlang.html#open_port-2

PR #689, opened 2022-01-20
Support rc versions of OTP; move unit tests

fhunleth:
Prerelease versions and OTP versions with 5 parts weren't being parsed.
This adds support for those and moves the tests to an easier-to-find
place.


lib/mix/nerves/utils.ex
```diff
@@ -160,33 +160,54 @@ defmodule Mix.Nerves.Utils do
...
-
-      # unknown
-      true ->
-        {:error, "Unable to Version.parse #{inspect(vsn)}"}
+  @doc """
+  Parse OTP versions
+
+  OTP versions can have anywhere from 2 to 5 parts. Normalize this into
+  a 3-part version for convenience. This is a lossy operation, but it
+  doesn't matter the the checks needed in this project.
```
jjcarstens, line 168:
```suggestion
  doesn't matter because the checks aren't needed in this project.
```
I think this is what you meant?

jjcarstens (approved):

PR #708, opened 2022-02-24
Lock mime ~> 1.6

jjcarstens:
This fixes #656 and #703 dependabot updates.

`mime ~> 2.0` requires Elixir >= 1.10, but Nerves supports >= 1.9. `plug`, `plug_cowboy`, and `mime` are only used during unit tests so just locking `mime` here will keep things moving along for now until the tests are adjusted or mime requires us to update


fhunleth:
Thanks!!!

PR #719, opened 2022-04-03
Adding installation instructions for NixOS and Nix package manager

jasonmj:
I've narrowed and tested the packages required to build firmware with Nerves. This brief description covers how to use them with `nix-shell` (on any linux based system with Nix package manager) or by including them in the `configuration.nix` file of a NixOS system. 

I tested that the build inputs work as expected on a clean nix instance on replit.com to make sure none of my host system's installed packages could cause false positives. 

Hopefully, this description makes sense and adds value to the Nerves installation docs. Please feel free to ask questions or make suggestions for changes. 

One thing I wasn't sure about, which you may want to change, is the addition of a heading above the note "For other linux host distributions...". It seems like it's easily lost as it blends with the text from the section I added.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -102,6 +102,46 @@ need to build a custom system. On Debian and Ubuntu, run the following:
 sudo apt install libssl-dev libncurses5-dev bc m4 unzip cmake python
 ```
 
+### For NixOS or Nix package manager
+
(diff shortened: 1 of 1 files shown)


fhunleth:
Thanks! Also, yes, I agree that the "For other Linux host distributions" needs to be made more visible now. I'll take a look.

PR #722, opened 2022-04-12
artifact.ex: Print string errors instead of &inspect/1

KoviRobi:
When the error message is a string, calling `inspect` on it limits it to
50 chars, which often ends up being something like
```
"[output of configure]" <> ...
```
and doesn't show the actual error.


fhunleth:
Thanks!

PR #724, opened 2022-04-27
add ssl parameter depth 3 to http_opts

mw23:
We are seeing issue with TLS fetching from S3 in CI with the following error:
```
21:26:41.176 [info]  TLS :client: In state :certify at ssl_handshake.erl:1952 generated CLIENT ALERT: Fatal - Handshake Failure
 - {:bad_cert, :max_path_length_reached}
 ```

Setting the ssl `depth` parameter to `3` like recommended in https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl fixes the issue.

### DIFF
--- a/lib/nerves/utils/http_client.ex
+++ b/lib/nerves/utils/http_client.ex
@@ -65,6 +65,7 @@ defmodule Nerves.Utils.HTTPClient do
         ssl: [
           verify: :verify_peer,
           cacertfile: CAStore.file_path(),
+          depth: 3,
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):

fhunleth:
Thanks for the PR. It looks good to me and I don't see why this would break anything. Give us a few days to try this out a little bit and then we'll make an official release with the update.

mw23:
Great - thanks for looking at it so quickly.

PR #731, opened 2022-05-04
Simplify fwup version check and fix dialyzer warning

fhunleth:


lib/mix/nerves/preflight.ex
```diff
@@ -40,33 +40,32 @@ defmodule Mix.Nerves.Preflight do
...
-        Please open a bug report for this issue on github.com/nerves-project/nerves
-        """)
+      You are running #{vsn}.
+      Please see https://hexdocs.pm/nerves/installation.html#fwup
+      for installation instructions
+      """)
     end
+
+    :ok
+  rescue
```
jjcarstens, line 62: If we raise above, then we're going to capture it here and then raise again

fhunleth: Thanks. I updated with a more minor change that gets past the dialyzer warning.

PR #741, opened 2022-05-05
nerves_bootstrap automatically calls add_alias

fhunleth:
These aren't needed, so delete all calls.


fhunleth:
Let me look at this some more. The tests pass if I remove the `Application.start(:nerves_bootstrap)` that I thought was getting called. I'm guessing that the `MIX_ENV=test` is causing `:nerves_bootstrap` to be implicitly started and all is fine, but I want to verify before merging for my own sanity.

jjcarstens:
This works because all the tests `use NervesTest.Case` which have a `load_env` function that calls `Application.ensure_all_started(:nerves_bootstrap)`

https://github.com/nerves-project/nerves/blob/main/test/support/test_case.ex#L142

PR #757, opened 2022-06-06
Refactor `Nerves.Release` and add test

jjcarstens:
This is the start of changes towards enabling credo. This file was one of the worst and needed much adjusting to satisfy the checks and become more readable.

It adjusts the ordering of the `roots.priorities` a bit, but it seemed like the order didn't fully matter (Just that beam came before app files which comes before private paths). The first commit is the test added without any code changes so that scripts can be compared by someone who knows more than me with this ordering


jjcarstens:
Well...looks like the test is pretty tied to the elixir version and Neves deps. Will need to think about how to make this much less brittle. It at least runs locally if the ordering needs to be inspected more

fhunleth:
Yes, this test is definitely tightly coupled to Erlang and Elixir versions. 

The beam files within an app are sorted in the opposite order that they're loaded in `shoehorn.script`. For example, in `shoehorn.script`, there's:

```
     {primLoad,
         ['Elixir.Access','Elixir.Agent','Elixir.Agent.Server',
          'Elixir.Application','Elixir.ArgumentError',
          'Elixir.ArithmeticError','Elixir.Atom','Elixir.BadArityError',
          'Elixir.BadBooleanError','Elixir.BadFunctionError',
          'Elixir.BadMapError','Elixir.BadStructError','Elixir.Base',
          'Elixir.Behaviour','Elixir.Bitwise','Elixir.Calendar',
          'Elixir.Calendar.ISO','Elixir.Calendar.TimeZoneDatabase',
```

but the priorities file has: 

```
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.Access.beam 31891
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.Agent.beam 31890
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.Agent.Server.beam 31889
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.Application.beam 31888
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.ArgumentError.beam 31887
srv/erlang/lib/elixir-1.13.4/ebin/Elixir.ArithmeticError.beam 31886
```

This "should" be the worst performing load order, but there are probably enough variables to be unsure about this. The intention is to lay the files out on the squashfs in the same order that they're read.

fhunleth:
@jjcarstens I rebased and fixed a Dialyzer warning. Still look good to you?

fhunleth:
Oh, it's failing CI.

jjcarstens:
We might need to just take the test out - I don't know how to make it consistent at this point. It works on my machine ¯\_(ツ)_/¯

jjcarstens:
nvm. This is ready now - When in doubt, exclude the release test from everything except the latest OTP 🤣

PR #763, opened 2022-07-04
Fix "Elixir Nx" hyperlink in README.md

RuiL1904:


fhunleth:
Oh wow, thanks for the fix!

PR #767, opened 2022-07-17
Improve Systems Compatibility section with per-system charts

mnishiguchi:
This pull request adds per-system compatibility charts to Systems documentation.

The HTML can be generated by running `scripts/nerves_system_compatibility.exs` based on relevant Github repos.


fhunleth:
Nice!

PR #774, opened 2022-08-07
Support using Elixir 1.14

fhunleth:


fhunleth:
I'm not sure what the best fix is here for the test error, but I really like how Elixir prints versions now.

PR #775, opened 2022-08-11
Update "Getting Started" page (top half)

mnishiguchi:
### Description

Large part of the "Getting Started" page is 3-6 years old as of 2022. While most information there is still useful, we could polish it for relevance to current Nerves and audience.

This pull request focuses on the top half of the "Getting Started" page. I kept the changes minimal.

The changes I made is totally subjective. I would like input from you all.

### Changes

- move mix command snippets closer to where they are explained about
- move extra `mix firmware.burn` options to its doc
  - reason: details can be noise from the beginners perspective
- replace `rpi3` with `rpi0` in snippets 
  - reason: assumption that many Nerves enthusiasts use `rpi0` these days
- clean code snippets so that they can be copy-paste'd easily

### DIFF
--- a/lib/mix/tasks/firmware.burn.ex
+++ b/lib/mix/tasks/firmware.burn.ex
@@ -20,6 +20,11 @@ defmodule Mix.Tasks.Firmware.Burn do
 
     * `--verbose` - produce detailed output about release assembly
 
+    * The `mix firmware.burn` task uses the `fwup` tool internally; any extra
+      arguments passed to it will be forwarded along to `fwup`. You can read
+      about the other supported options in the
+      [`fwup` documentation](https://github.com/fwup-home/fwup#invoking).
+
   ## Environment variables
 
     * `NERVES_SYSTEM`    - may be set to a local directory to specify the Nerves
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/firmware.burn.ex
```diff
@@ -32,7 +37,14 @@ defmodule Mix.Tasks.Firmware.Burn do
...
   ```
   # Upgrade the contents of the SDCard at /dev/mmcblk0 using the rpi0 system
-  NERVES_SYSTEM=rpi0 mix firmware.burn --device /dev/mmcblk0 --task upgrade
+  mix firmware.burn --device /dev/mmcblk0 --task upgrade
+  ```
+
+  If you are sure there is only one SD card inserted, you can also add the `-y`
+  flag to skip the confirmation that it is the correct device.
+
+  ```
```
fhunleth, line 46:
```suggestion
  ```bash
```

fhunleth (approved):

fhunleth:
Nice! You're right. This has gotten a bit dated. 

I'm actually not sure what people use most often these days since all of the Raspberry Pis are so hard to find. Switching to the `rpi0` makes sense, though.

fhunleth:
Thanks!

PR #779, opened 2022-08-23
Adjust release test for Elixir 1.14.0

jjcarstens:


jjcarstens:
This is a start. This test is super brittle when we change Elixir versions. Maybe we need a expected files based on the Elixir version detected?

jjcarstens:
Adjusted to just check the first few lines that should be consistent between 1.13 and 1.14 versions. It makes the test less brittle for now, but also even less accurate. We can revisit when we work on the ordering optimization

fhunleth:
Sounds good.

PR #782, opened 2022-09-03
Use Elixir 1.14.0 throughout

fhunleth:


docs/Installation.md
```diff
@@ -192,10 +192,10 @@ asdf plugin-add elixir
 # latest official Nerves systems are compatible with the versions below. In
 # general, differences in patch releases are harmless. Nerves detects
 # configurations that might not work at compile time.
-asdf install erlang 24.3.2
-asdf install elixir 1.13.3-otp-24
-asdf global erlang 24.3.2
-asdf global elixir 1.13.3-otp-24
+asdf install erlang 25.0.3
```
jjcarstens, line 195:
```suggestion
asdf install erlang 25.0.4
```

docs/Installation.md
```diff
@@ -192,10 +192,10 @@ asdf plugin-add elixir
...
+asdf install elixir 1.14.0-otp-25
+asdf global erlang 25.0.3
```
jjcarstens, line 197:
```suggestion
asdf global erlang 25.0.4
```

jjcarstens (changes requested):
Looks like something isn't right with the CI config

docs/Installation.md
fhunleth: I really don't think it matters, but the systems are still on 24.0.3.

fhunleth: I'm updating `nerves_system_br` to 25.0.4. PR should be up shortly...

jjcarstens (approved):

PR #785, opened 2022-09-10
Updating OTP version of Elixir install to 25

alanj853:
When following the 'Getting Started' documentation as is, you will get a host mismatch error when using the 'rpi0' target, as https://github.com/nerves-project/nerves_system_rpi0 uses OTP 25 since v1.19.0

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -193,9 +193,9 @@ asdf plugin-add elixir
 # general, differences in patch releases are harmless. Nerves detects
 # configurations that might not work at compile time.
 asdf install erlang 24.3.2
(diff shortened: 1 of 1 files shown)


fhunleth:
Thanks for catching this! The line above the Elixir ones that you changed also needs to be updated to install Erlang/OTP 25. I'll do that in a subsequent commit and update hex.

alanj853:
> Thanks for catching this! The line above the Elixir ones that you changed also needs to be updated to install Erlang/OTP 25. I'll do that in a subsequent commit and update hex.

Sorry, forgot to add them in the commit.

PR #789, opened 2022-09-17
Add a tip of deleting cached artifacts

mnishiguchi:
### Description

Nerves automatically caches Nerves artifacts that are downloaded for our Nerves projects for our convenience. As time progresses, it may result it a pile of old unused Nerves artifacts in our development host machine.

How about show the Nerves users how to delete those archives?

### Changes

Add a pro tip box with a brief explanation


fhunleth:
Looks good. Thanks!

PR #790, opened 2022-09-17
Adjust release test to better support version changes

jjcarstens:
Asserting a specific generated priorities is brittle as it drastically changes with each Elixir version. The ordering is also currently suboptimal.

For now, just check the file begins with the consolidated protocols until the ordering is optimized and a better test can be constructed


jjcarstens:
Oops. Accidental duplicate

PR #792, opened 2022-09-17
Check if supplied rootfs_overlays have incompatible directories

jjcarstens:
Fixes https://github.com/nerves-project/nerves_system_br/issues/495

Some directories are mounted as filesystems on the device which overwrite any files the overlay may have placed there. To prevent confusion, this will detect these clashes and fail the firmware build with a helpful warning.

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -112,6 +112,8 @@ defmodule Mix.Tasks.Firmware do
           [Path.expand(overlay)]
       end
 
+    prevent_overlay_overwrites!(project_rootfs_overlay)
+
     rootfs_overlays =
       [build_rootfs_overlay | project_rootfs_overlay]
       |> Enum.map(&["-a", &1])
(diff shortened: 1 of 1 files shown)


mnishiguchi (approved):
✅ 

target: `rpi0`

in `mix.exs`

```elixir
...
{:nerves, runtime: false, github: "nerves-project/nerves", branch: "overlay-check", override: true},
...
```

in `rootfs_overlay`



Running `mix firmware`, I got this error message as expected



I deleted `rootfs_overlay/data/hello.txt` and re-ran `mix firmware` and the build succeeded.

lib/mix/tasks/firmware.ex
```diff
@@ -112,6 +112,8 @@ defmodule Mix.Tasks.Firmware do
           [Path.expand(overlay)]
       end
 
+    _ = prevent_overlay_overwrites!(project_rootfs_overlay)
```
fhunleth, line 115:
```suggestion
    prevent_overlay_overwrites!(project_rootfs_overlay)
```

Move the ignored return value to inside the function. I don't know what this would return.

lib/mix/tasks/firmware.ex
```diff
@@ -263,4 +265,28 @@ defmodule Mix.Tasks.Firmware do
...
+      The firmware contains overlay files which reference directories that are
+      mounted as file systems on the device. The filesystem mount will completely
+      overwrite the overlay and these files will be lost.
+
+      Remove the following overlay directories and build the firmware again:
+
+      #{for dir <- shadow_mounts, do: "  * #{dir}\n"}
+      #{IO.ANSI.reset()}https://hexdocs.pm/nerves/advanced-configuration.html#root-filesystem-overlays
+      """)
+    end
```
fhunleth, line 290: I suspect that Dialyzer is going to want something slightly more than this...

```suggestion
    end
    
    :ok
```

jjcarstens: Seems dialyzer didn't care if this was there or not, so left it out

fhunleth:
This is going to save people so much time. Thanks for adding.

PR #797, opened 2022-09-28
Remove deprecated Mix.Config

mnishiguchi:
[Mix.Config](https://hexdocs.pm/mix/Mix.Config.html) has long been deprecated. Let's use [Config](https://hexdocs.pm/elixir/Config.html) instead per Elixir doc.

Also, I removed kind of redundant `import Config` in our example config snippets. I thought it was confusing since we do not need to add it to the existing file. `import Config` is so idiomatic that the user will figure out.


jjcarstens (approved):

fhunleth:
Wow, thanks for fixing that!

PR #803, opened 2022-10-23
Add Nerves cheatsheet

fhunleth:
This needs some work, but it seems really useful to have one.


docs/Nerves Cheatsheet.cheatmd
```diff
@@ -0,0 +1,160 @@
...
+Write the most recently built firmware to MicroSD:
+
+```sh
+mix burn
+```
+
+### Update firmware via ssh
+
+```sh
+mix firmware.push [hostname or IP address]
```
jjcarstens, line 64:
```suggestion
mix upload [hostname or IP address]
```
Is this still I think? I thought we removed `firmware.push`

jjcarstens (commented):
Seems weird to include pieces of VintageNet and tool shed when they are technically dependencies that could be excluded in nerves.  But also...when do you not include those? ¯\_(ツ)_/¯

docs/Nerves Cheatsheet.cheatmd
fhunleth: lol - I was copy/pasting from an old Nerves PDF cheatsheet and missed that.

fhunleth:
> Seems weird to include pieces of VintageNet and tool shed when they are technically dependencies that could be excluded in nerves. But also...when do you not include those? ¯_(ツ)_/¯

This is where I'm somewhat confused as to what the ex_doc cheatsheet is for. I had thought that the intention was that you'd have the cheatsheet in a browser window on a second monitor and look back at it when you're using the library. You wouldn't have to scroll much since the information was compactly presented. For Nerves, I took that as the highlights to get a device working with Nerves. I always include networking (mostly WiFi), logging and iex prompt use as well, since literally everyone has questions about those in the beginning. I don't think there needs to be much detail, but I can provide examples from training and slack support for why I included each. If the goal is to fit everything on the screen at the same time, I can go back and prioritize. 

Having said that, I was thinking that other Nerves-related libraries may be easier. Delux actually seems like it might be a good one. Grizzly does too since its cookbook is basically a cheatsheet.

PR #819, opened 2023-01-01
Add Nerves basics cheatsheet

mnishiguchi:
A prototype of Nerves basics cheatsheet

### DIFF
--- a/cheatsheets/basics.cheatmd
+++ b/cheatsheets/basics.cheatmd
@@ -1,0 +1,189 @@
+# Nerves basics
+
+## Terminology
+
+| Term            | Definition                                                                                                            |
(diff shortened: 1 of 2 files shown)


fhunleth:
This looks like a great start. Thanks for make it. 

Do you know what's up with the left pane in your screenshots?

mnishiguchi:
> Do you know what's up with the left pane in your screenshot?

The screenshot is kind of artificially made while scrolling so that the entirety can be captured. With regard to the left pane, the net change would be the cheatsheets group appended to the bottom. The grouping is optional. I believe cheatsheets can coexist with regular markdown documents if we wish.

PR #822, opened 2023-01-31
Refactor GitHub artifact fetching

jjcarstens:
* Switches `:github_releases` artifact site to also use `Resolvers.GithubAPI`
* `Resolvers.GithubAPI` has been totally refactored
  * Fixed issue with errors reporting to use `GITHUB_TOKEN`, but that env var would never actually be used in the code
  * Supports `GITHUB_TOKEN` or the `GH_TOKEN` shorthand that GitHub also supports
  * Adjusted errors to better report the situation discovered
    * No release
    * Release found but no artifacts * Artifacts found, but none with matching checksum * Missing token when unable to authenticate a GitHub request to private repo

## No valid checksum
image

## No release found
image

## Missing token
image

## No artifacts in release
image

### DIFF
--- a/lib/nerves/artifact/resolvers/github_api.ex
+++ b/lib/nerves/artifact/resolvers/github_api.ex
@@ -2,143 +2,132 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
   @moduledoc false
   @behaviour Nerves.Artifact.Resolver
 
-  alias Nerves.Utils
-  alias Nerves.Utils.HTTPClient
+  alias Nerves.{Utils, Utils.HTTPClient}
 
   @base_url "https://api.github.com/"
-  @required_opts [:username, :token, :tag]
+
+  defstruct artifact_name: nil,
+            base_url: @base_url,
+            headers: [],
+            http_client: HTTPClient,
+            http_pid: nil,
+            public?: false,
+            opts: [],
+            repo: nil,
+            tag: "",
+            token: "",
+            url: nil,
+            username: ""
 
   @impl Nerves.Artifact.Resolver
   def get({org_proj, opts}) do
-    case validate_opts(opts) do
-      {:ok, opts} ->
-        fetch_artifact(org_proj, opts)
+    opts =
+      %{struct(__MODULE__, opts) | opts: opts, repo: org_proj}
+      |> maybe_adjust_token()
+      |> add_http_opts()
+      |> maybe_start_http()
 
-      {:error, missing} ->
-        missing = Enum.join(missing, ", ")
-        artifact_name = opts[:artifact_name]
+    result = fetch_artifact(opts)
 
-        error = """
-        Skipping: #{artifact_name}
-        Required options missing: [#{missing}]
-        """
+    opts.http_client.stop(opts.http_pid)
 
-        {:error, error}
-    end
-  end
-
-  defp fetch_artifact(org_proj, opts) do
-    artifact_name = opts[:artifact_name]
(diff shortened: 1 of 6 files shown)


lib/nerves/artifact/resolvers/github_api.ex
```diff
@@ -2,143 +2,132 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
...
+            """
+            Missing token
 
-        For example:
+                 For private releases, you must authenticate the request to fetch release assets.
+                 You can do this in a few ways:
 
-        export GITHUB_TOKEN=<your_token>
+                   * export or set GITHUB_TOKEN=<your-token>
+                   * set `token: <your-token>` for this artifact site in your Nerves system mix.exs
```
fhunleth, line 99:
```suggestion
                   * set `token: <get-token-function>` for this GitHub repository in your Nerves system mix.exs
```

Since we don't want people storing tokens in their `mix.exs`, I changed the line to give a hint to use a function. 

I also changed artifact site to GitHub repository, since I think that more directly corresponds to what they need from GitHub. There might be a better word?

jjcarstens: Ya, I was struggling on how best to word this to direct users to their `nerves_system -> mix.exs -> nerves_package/0 -> :artifact_sites` setting. I think this change is better

lib/nerves/artifact/resolvers/github_api.ex
```diff
@@ -2,143 +2,132 @@ defmodule Nerves.Artifact.Resolvers.GithubAPI do
...
+      | headers: headers,
+        url: Path.join([opts.base_url, "repos", opts.repo, "releases", "tags", opts.tag])
+    }
   end
 
-  @spec validate_opts(Keyword.t()) :: {:ok | :error, Keyword.t()}
-  def validate_opts(opts) do
-    keys = Keyword.keys(opts)
+  defp maybe_adjust_token(opts) do
+    token = System.get_env("GITHUB_TOKEN") || System.get_env("GH_TOKEN")
```
amclain, line 54: > Supports `GITHUB_TOKEN`

FYI as I've set up Nerves projects with GitHub Actions instead of CircleCI, it's prevented storing secrets that start with `GITHUB` because it's a reserved word. I got around this by changing the secret to something like `ARTIFACT_TOKEN` and updating the CI config. Just thought I'd mention it while I see this code changing. Also maybe that's what `GH_TOKEN` is for and I just didn't see it earlier...

jjcarstens: Interesting. `GH_TOKEN` was to match the [GitHub CLI supported environment variables](https://cli.github.com/manual/gh_help_environment). I thought `GITHUB_*` would take precedence, but looks like it is the other way around. We could switch that.

I would say that we leave it as is rather than try to make up new environment variables. This is more of a convenience and to potentially provide more helpful errors. If some special env needs to be used, I would say that should go in your Nerves systems `nerve_package()` function when defining the artifact:

```elixir
def nerves_package() do
  [
    artifact_sites: [
      {:github_api, "my_org/my_system/, token: System.get_env("MY_ARTIFACT_TOKEN")}
    ]
  ]
end
```

fhunleth (approved):

lib/nerves/artifact/resolvers/github_api.ex
amclain: >I would say that we leave it as is

👍 Sure, no problem. I've been using a workaround similar to what you posted and it was also easy enough to update the CI config for it.

PR #823, opened 2023-01-31
Remove duplicate artifact request

jjcarstens:
In the very beginning, artifacts were stored in releases with their full checksum value. Now they are stored with the 7 byte checksum shorthand, but we still try to request both.

I think its been long enough using the shorthand that we can remove this second request which should clear up the logs a bit and make it less confusing when an artifact can't be found

Also, the work in #822 should provide better output when artifacts can't be found which would help as well


fhunleth:
I agree. The checksum in the filename is really only to catch incorrect archive issues very quickly. 7 digits is more than enough for that.

fhunleth (approved):

PR #828, opened 2023-02-10
Add missing pkg-config dependency to shell.nix

matteo-briani:
Add `pkg-config` dependency to the `shell.nix`.
pkg-config is needed to build the firmware.


jjcarstens (approved):

jjcarstens:
Thanks!

PR #833, opened 2023-02-16
Remove :json_codec configuration option 

jjcarstens:
Since `Jason` is now a required dependency, this removes the ability to specify alternate JSON libraries and emits a warning if is found in the Application config environment

### DIFF
--- a/lib/nerves/utils.ex
+++ b/lib/nerves/utils.ex
@@ -2,7 +2,9 @@ defmodule Nerves.Utils do
   @moduledoc false
   @alphanum 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_'
 
-  @json_codec Jason
+  if Application.compile_env(:nerves, :json_codec) do
+    IO.warn(":json_codec is no longer supported. Please remove from your config", [])
+  end
 
   <<i1::32-unsigned-integer, i2::32-unsigned-integer, i3::32-unsigned-integer>> =
     :crypto.strong_rand_bytes(12)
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -38,7 +38,7 @@ defmodule Nerves.MixProject do
     [
       {:castore, "~> 0.1"},
       {:elixir_make, "~> 0.6", runtime: false},
-      {:jason, "~> 1.2", optional: true},
+      # {:jason, "~> 1.2", optional: true},
```
fhunleth, line 41: Why is the `:jason` dependency commented out?

lib/nerves/utils.ex
```diff
@@ -2,7 +2,18 @@ defmodule Nerves.Utils do
...
+  @json_codec Application.compile_env(:nerves, :json_codec) || Jason
+
+  unless Code.ensure_loaded?(@json_codec) do
+    Mix.raise("""
+    Nerves requires a JSON codec to be defined.
+
+    Please include :jason as a dependency or configure your own JSON parser
+    by updating your config.exs
+
+      config :nerves, json_codec: Poison
```
fhunleth, line 14:
```suggestion
      config :nerves, json_codec: Jason
```

mix.exs
jjcarstens: O derp 🤦 because i commented out in order to test the error happens

lib/nerves/utils.ex
jjcarstens: Jason is the default. This is just showing that it could be something different. The message could probably be updated entirely...

jjcarstens:
@fhunleth Are we happy with the new error message here?

fhunleth:
Isn't this moot since Jason is a required dependency now?

jjcarstens:
Mostly, but we still allow use to specify an alternate JSON codec. This is catching that case that they configure something bad.

Though it could probably be adapted to make sure it has decode and encode functions in order to be a fully complete check

fhunleth:
How about we don't allow users to specify a JSON codec anymore?

fhunleth (approved):

PR #837, opened 2023-02-26
Prevent accidentally installing nerves as an archive

fhunleth:
This doesn't work and is surprising confusing since the error happens
long after the archive.install step.


mnishiguchi (approved):
Looks great 🎉 

CleanShot 2023-02-26 at 07 59 12

PR #841, opened 2023-03-03
Use GitHubAPI for public release artifacts

jjcarstens:
Brings back the use of GitHubAPI for public release artifacts.

Also adds fallback to public URL for public release when the API rate limit is reached rather than just failing. If it still fails, the error message won't be as useful. But that is worth it if it succeeds

### DIFF
--- a/lib/nerves/artifact.ex
+++ b/lib/nerves/artifact.ex
@@ -365,21 +365,13 @@ defmodule Nerves.Artifact do
   defp expand_site(_, _, _ \\ [])
 
   defp expand_site({:github_releases, org_proj}, pkg, opts) do
-    # TODO: Switch to GithubAPI once JSON codec expectations are
-    # fixed within the lib
-    # opts =
-    #   opts
-    #   |> Keyword.put(:artifact_name, download_name(pkg, opts) <> ext(pkg))
-    #   |> Keyword.put(:public?, true)
-    #   |> update_in([:tag], &(&1 || "v#{pkg.version}"))
-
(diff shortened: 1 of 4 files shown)


lib/nerves/artifact.ex
```diff
@@ -365,21 +365,13 @@ defmodule Nerves.Artifact do
...
-    # {Resolvers.GithubAPI, {org_proj, opts}}
-
-    expand_site(
-      {:prefix, "https://github.com/#{org_proj}/releases/download/v#{pkg.version}/"},
-      pkg,
+    opts =
       opts
-    )
+      |> Keyword.put(:artifact_name, download_name(pkg, opts) <> ext(pkg))
+      |> Keyword.put(:public?, true)
```
fhunleth, line 371: Should `:public?` be forced to `true`? How do private GitHub repos work with this?

jjcarstens: The current behavior has been that `:github_Release` was used for public releases downloaded from a public URL. To use a private release, one would use `:github_api` with a token

The changes here are that both use the GitHubAPI resolver so that helpful error messages can be made when something goes wrong, but forces `github_release` to be some public call as usual

That said, this could prob be changed to be a `put_new` just so that it is overridable

fhunleth: I see. I suppose that I don't have strong feelings. Maybe at some point we unify this, but keeping everything working sounds good for now. We need to make a patch release in the near term anyway.

fhunleth (approved):

PR #842, opened 2023-03-04
Move compatibility script downloads to tmp folder 

jjcarstens:
The compatibility script would download everything to `scripts/tmp` which could cause issues when trying to publish this package since all those files would be included according to the `mix.exs`.

This changes to use `System.tmp_dir!` for the download files and also adds a cleanup step at the end to remove the bulk of the content


fhunleth:
How about updating the script that creates that `tmp` directory to put it under `System.tmp` and clean it up?

jjcarstens:
@fhunleth Updated to use tmp dir and do some cleanup

PR #843, opened 2023-03-04
Allow castore v1.0 to be used

fhunleth:


fhunleth:
`elixir_make` needs to be released to trigger `castore` to actually update to v1.0.0, but I imagine this will happen sometime soon.

jjcarstens (approved):

mnishiguchi (approved):

PR #850, opened 2023-03-13
Add `compile_fixture` to `NervesTest.case`

jjcarstens:
Many tests currently rely on `File.cd!/2`, but this [changes the directory globally](https://hexdocs.pm/elixir/File.html#cd!/2) causing race conditions and flaky tests.

This allows us to separate out fixture setup and runs by process by only managing the setup and letting the test process use safe directory changes like with `System.cmd`

This also updates the easier tests to use this new function as a starting point

### DIFF
--- a/test/fixtures/integration_app/config/config.exs
+++ b/test/fixtures/integration_app/config/config.exs
@@ -1,0 +1,3 @@
+import Config
+
+Application.start(:nerves_bootstrap)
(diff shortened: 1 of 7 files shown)


jjcarstens:
It needs https://github.com/nerves-project/nerves/pull/851 so we are _actually_ bootstrapping as expected during tests

jjcarstens:
victory!

PR #851, opened 2023-03-13
Remove `nerves_bootstrap` mix dependency

jjcarstens:
Attempting to use `nerves_bootstrap` in this way currently as no effect. The archive is needed to bootstrap in and will always look to the system for it to be installed.

This has gone relatively unnoticed because all of the integration tests are manually manipulating `Mix` API and forcing bootstrap behavior rather than relying on the installed archive as expected


mnishiguchi (approved):
Make sense to me.

jjcarstens:
I don't think I'm going to get CI to work until I get all the rest of the changes. Here's some proof things are still okay

image

fhunleth (approved):
Sounds good. Thanks for taking the time to fix the tests once and for all.

Issue #880, opened 2023-06-12
Formatting in installation instructions are unclear


dpurrington:
### Environment

n/a

### Current behavior

[These instructions](https://hexdocs.pm/nerves/installation.html#linux) make it appear as if all the text under "For Arch based systems" applies only to Arch-based systems, when in fact they apply to all Linux systems. The result is Debian system users (and others) will not perform all the necessary installation tasks -- specifically installation of `fwup.`

### Expected behavior

It should be clear which tasks apply to all Linux systems, and which are specific to particular distributions.

mnishiguchi:
Thank you for raising this issue!

PR #882, opened 2023-06-12
Update Installation.md for clarity regarding fwup

dpurrington:
In the prior version it appeared as if fwup instructions were only for arch linux distributions, due to how the headings were arranged.

This addressed [issue 880](https://github.com/nerves-project/nerves/issues/880)

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -72,15 +72,17 @@ cards. It is important that you install `fwup ` in both environments.
 
 ## Linux
 
-First, install a few packages using your package manager:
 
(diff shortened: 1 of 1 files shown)


mnishiguchi (commented):
Thanks for addressing the readability of this document. I see some room for improvement.

I am wondering if this extra nesting is necessary because it seems like the `On NixOS or with Nix package manager` section contains `fwup`. But how can it be better... 🤔

mnishiguchi:
resolves https://github.com/nerves-project/nerves/issues/880

PR #884, opened 2023-06-30
Prevent bad vm.args in Nerves firmware

jjcarstens:
With Elixir 1.15, the way the Elixir IEx shell is started on device has changed and using the old way will break things.

This adds a new check during the release (firmware) initialization and stops the firmware build if the old method is still there

> **Note**
> Using Elixir 1.15 
> image
>
> Using Elixir < 1.15
> image

### DIFF
--- a/lib/nerves/release.ex
+++ b/lib/nerves/release.ex
@@ -174,6 +174,11 @@ defmodule Nerves.Release do
   @legacy_elixir_opts ["-user Elixir.IEx.CLI"]
   defp check_vm_args_compatibility!(release) do
     Mix.shell().info([:yellow, "* [Nerves] ", :reset, "validating vm.args"])
+    vm_args_path = Mix.Release.rel_templates_path(release, "vm.args.eex")
+
+    if not File.exists?(vm_args_path) do
+      Mix.raise("Missing required #{vm_args_path}")
+    end
 
     {exclusions, inclusions} =
       if Version.match?(System.version(), ">= 1.15.0") do
--- a/lib/nerves/release.ex
+++ b/lib/nerves/release.ex
@@ -16,6 +16,8 @@ defmodule Nerves.Release do
         steps: release.steps ++ [&Nerves.Release.finalize/1]
     }
 
+    check_vm_args_compatibility!(release)
+
     _ = File.rm_rf!(release.path)
 
     if Code.ensure_loaded?(Shoehorn.Release) do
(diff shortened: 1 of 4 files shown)


lib/nerves/release.ex
```diff
@@ -167,4 +169,58 @@ defmodule Nerves.Release do
...
+                "* ",
+                vm_args_path,
+                ":",
+                to_string(num),
+                "\n",
+                :red,
+                "\n  - ",
+                line,
+                :green,
+                "  + the_new_way\n",
```
jjcarstens, line 202: Prob shouldn't merge until we have this new way figured out

fhunleth: The new way is to replace this:

```
-noshell
-user Elixir.IEx.CLI
```

with this:

```
-noshell
-user elixir
-run elixir start_iex
```

fhunleth:
I just realized that we need the opposite check as well since someone might be using Elixir 1.14. I.e., if they specify the Elixir 1.15 way, give them an error.

jjcarstens:
Good point. Is this also specific to OTP 26?

fhunleth:
I don't believe so. I was using the old way with Elixir 1.14 and OTP 26. Unfortunately I feel like things in this area have been changing lately and checking again with Elixir 1.14.5 would be wise.

lib/nerves/release.ex
```diff
@@ -167,4 +169,82 @@ defmodule Nerves.Release do
...
+
+    if length(errors) > 0 do
+      errs = IO.ANSI.format(errors) |> IO.iodata_to_binary()
+
+      Mix.raise("""
+      Incompatible vm.args.eex
+
+      Nerves requires a few specific pieces in vm.args.eex to correctly start the BEAM
+      and other things like IEx on UART/HDMI. Incorrect or missing settings could
+      prevent your Nerves firmware from successfully booting.
```
fhunleth, line 205: How's this?

```suggestion
      The procedure for starting IEx changed in Elixir 1.15. The rel/vm.args.eex for
      this project starts IEx in an incompatible way for the version of Elixir you're
      using and won't work. 
      
      To fix this, either change the version of Elixir that you're using or make the
      following changes to vm.args.eex:
```

lib/nerves/release.ex
```diff
@@ -167,4 +169,82 @@ defmodule Nerves.Release do
...
+
+    vm_args = File.read!(vm_args_path)
+
+    errors =
+      []
+      |> check_vm_args_inclusions(vm_args, inclusions, vm_args_path)
+      |> check_vm_args_exclusions(vm_args, exclusions, vm_args_path)
+
+    if length(errors) > 0 do
+      errs = IO.ANSI.format(errors) |> IO.iodata_to_binary()
```
fhunleth, line 198:
```suggestion
      errs = IO.ANSI.format(errors) |> IO.chardata_to_string()
```

fhunleth (approved):
Looks good to me.

@mnishiguchi What do you think?

mnishiguchi (approved):

PR #886, opened 2023-07-03
Elixir 1.15 adjustments

jjcarstens:


fhunleth (approved):
This looks good to me aside from the Dialyzer warnings, but 👍🏻 after those are fixed.

PR #895, opened 2023-07-22
Bump Elixir version; test w/ old Alpine versions again

fhunleth:
While it seems unlikely, testing with older Alpine versions has caught
things in the past so bring them back.


jjcarstens (approved):

jjcarstens:
Looks like you already pushed these commits to main

fhunleth:
Weird. Normally GitHub automatically detects when I merge to `main` via the command line. I thought I made a mistake, but the hashes are the same between this branch and main. Whelp, I guess I close with comment.

PR #897, opened 2023-07-31
Adjust `mix nerves.system.shell` for OTP 26

jjcarstens:
OTP 26 made some big changes around serial interface and it breaks the `mix nerves.system.shell` task. In the end, the main issue is that we no longer have an easy way to take over the STDIN of the task to forward it to a shell running in the system build directory.

Many attempts were made to recapture STDIN, but decided that for now we need to move on. Instead, this adjusts the task to do _all_ the setup for the build directory as before and outputs a warning to the user for the command that can be manually run in order to get the same shell that was previously started with `mix nerves.system.shell`

**Docker error**
image

**Local (Linux) error**
image

### DIFF
--- a/lib/nerves/artifact/build_runners/docker.ex
+++ b/lib/nerves/artifact/build_runners/docker.ex
@@ -125,30 +125,62 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
 
   @doc """
   Connect to a system configuration shell in a Docker container
+
+  Unsupported in >= OTP 26. However, the Docker env will still be created
+  and a command output to IO which can be used manually.
   """
   @spec system_shell(Nerves.Package.t()) :: :ok
   def system_shell(pkg) do
     _ = preflight(pkg)
     {_, image} = config(pkg)
-    platform_config = pkg.config[:platform_config][:defconfig]
-    defconfig = Path.join("/nerves/env/#{pkg.app}", platform_config)
-
-    initial_input = [
-      "echo Updating build directory.",
-      "echo This will take a while if it is the first time...",
-      "/nerves/env/platform/create-build.sh #{defconfig} #{@working_dir} >/dev/null"
-    ]
 
     mounts = Enum.join(mounts(pkg), " ")
     ssh_agent = Enum.join(ssh_agent(), " ")
     env_vars = Enum.join(env(), " ")
 
-    cmd =
+    shell =
       "docker run --rm -it -w #{@working_dir} #{env_vars} #{mounts} #{ssh_agent} #{image} /bin/bash"
 
     set_volume_permissions(pkg)
 
-    Mix.Nerves.Shell.open(cmd, initial_input)
+    create_build = create_build_cmd(pkg) |> Enum.join(" ")
+
+    if String.to_integer(System.otp_release()) < 26 do
+      # Legacy shell support
+      initial_input = [
+        "echo Updating build directory.",
+        "echo This will take a while if it is the first time...",
+        "#{create_build} >/dev/null"
+      ]
+
+      Mix.Nerves.Shell.open(shell, initial_input)
+    else
+      Mix.Nerves.IO.shell_warn("shell start deprecated", """
+      OTP 26 made several changes to the serial interface handling. Unfortunately, this
+      is a regression in preventing the Nerves tooling from starting a system sub-shell.
+
+      However, the Docker environment has been configured and can be used the same as
+      before by manually running the command below which creates the build directory
+      and gives you a shell running in the container to interact with the system:
+      """)
+
+      exec_input = [
+        "echo -e '\\e[25F\\e[0J\\e[1;7m\\n  Preparing Nerves Shell  \\e[0m'",
+        "echo -e '\\e]0;Nerves Shell\\a'",
+        "echo \\\"PS1='\\e[1;7m Nerves \\e[0;1m \\W > \\e[0m'\\\" >> ~/.bashrc",
+        "echo \\\"PS2='\\e[1;7m Nerves \\e[0;1m \\W ..\\e[0m'\\\" >> ~/.bashrc",
+        "echo 'Updating build directory.'",
+        "echo 'This will take a while if it is the first time...'",
+        "#{create_build} >/dev/null",
+        "echo -e '\\nUse \\e[33mctrl+d\\e[0m or \\e[33mexit\\e[0m to leave the container shell\\n'",
+        "exec /bin/bash"
+      ]
+
+      # >= OTP 26 will just output this command to the user for them to run
+      # With Docker, we want create-build.sh to run and then to exec the shell
+      # to keep the docker container open as an interactive shell
+      Mix.shell().info(~s(#{shell} -c "#{Enum.join(exec_input, " ; ")}"))
+    end
   end
 
   defp preflight(pkg) do
@@ -162,10 +194,14 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
 
   # Build Commands
 
-  defp create_build(pkg, stream) do
+  defp create_build_cmd(pkg) do
     platform_config = pkg.config[:platform_config][:defconfig]
     defconfig = Path.join("/nerves/env/#{pkg.app}", platform_config)
-    cmd = ["/nerves/env/platform/create-build.sh", defconfig, @working_dir]
+    ["/nerves/env/platform/create-build.sh", defconfig, @working_dir]
+  end
+
+  defp create_build(pkg, stream) do
+    cmd = create_build_cmd(pkg)
     shell_info("Starting Build... (this may take a while)")
     run(pkg, cmd, stream)
   end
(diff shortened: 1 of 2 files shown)


fhunleth:
What do you do to continue the build after you run Docker?

jjcarstens:
The output docker command is the last step of "starting the system shell" - You can do the typical things there such as `make menuconfig` or just use `make` to compile. Everything is referenced in a build volume.

Later one, if one uses the system as a path dependency then those same build artifacts are used with `mix compile` and `mix firmware`

lib/mix/tasks/nerves.system.shell.ex
```diff
@@ -70,11 +65,21 @@ defmodule Mix.Tasks.Nerves.System.Shell do
...
+      OTP 26 made several changes to the serial interface handling. Unfortunately, this
+      a regression in preventing the Nerves tooling from starting a system sub-shell.
 
-    # Set :user back to the real one
-    Process.register(user, :user)
+      However, the build directory and all needed pieces have been created and can be
+      accessed by manually running the command below:
 
-    :ok
+      #{[IO.ANSI.reset(), cmd]}
```
fhunleth, line 81: Is this how the workflow works?

```suggestion
      #{[IO.ANSI.reset(), cmd]}
      
      When you're in the container, run `make` to build the system as is or run `make menuconfig`
      to modify it. When done, exit the container and run `mix firmware` again.
```

jjcarstens:
@fhunleth This is ready for another review. I adjusted it to output the usable command at the end. A really hacky workaround is a user could do `eval "$(mix nerves.system.shell | tail -n 1)"` to get the same effect in OTP 26, or just copy/paste the output command in the error.

fhunleth (approved):

guillego:
This is a great improvement! Not the smooth experience of OTP 25 but much better than the cryptic error message.

The only caveat I have is this message after running the long docker command:

```
docker run --rm -it -w /nerves/build --env UID=501 --env GID=20 --mount type=volume,src=nerves_system_rpi0_video-iJFrgQHhlZ5Yv6f8,target=/nerves/build --mount type=bind,src=/Users/guille/.nerves/dl,target=/nerves/dl --mount type=bind,src=/Users/guille/personal/nerves_system_rpi0_video,target=/nerves/env/nerves_system_rpi0_video --mount type=bind,src=/Users/guille/personal/nerves_system_rpi0_video/deps/nerves_system_br,target=/nerves/env/platform --env NERVES_BR_DL_DIR=/nerves/dl -v /private/tmp/com.apple.launchd.R14n1eLnGE/Listeners:/ssh-agent -e SSH_AUTH_SOCK=/ssh-agent ghcr.io/nerves-project/nerves_system_br:1.23.3 /bin/bash -c "/nerves/env/platform/create-build.sh /nerves/env/nerves_system_rpi0_video/nerves_defconfig /nerves/build ; exec /bin/bash"
 guille@calsot  ~/personal/nerves_system_rpi0_video   main ±  docker run --rm -it -w /nerves/build --env UID=501 --env GID=20 --mount type=volume,src=nerves_system_rpi0_video-iJFrgQHhlZ5Yv6f8,target=/nerves/build --mount type=bind,src=/Users/guille/.nerves/dl,target=/nerves/dl --mount type=bind,src=/Users/guille/personal/nerves_system_rpi0_video,target=/nerves/env/nerves_system_rpi0_video --mount type=bind,src=/Users/guille/personal/nerves_system_rpi0_video/deps/nerves_system_br,target=/nerves/env/platform --env NERVES_BR_DL_DIR=/nerves/dl -v /private/tmp/com.apple.launchd.R14n1eLnGE/Listeners:/ssh-agent -e SSH_AUTH_SOCK=/ssh-agent ghcr.io/nerves-project/nerves_system_br:1.23.3 /bin/bash -c "/nerves/env/platform/create-build.sh /nerves/env/nerves_system_rpi0_video/nerves_defconfig /nerves/build ; exec /bin/bash"
Setting User
UID: 501
GID: 20
Switching user
make: Entering directory '/nerves/env/platform/buildroot-2023.02.3'
  GEN     /nerves/build/Makefile

WARNING: unmet direct dependencies detected for BR2_PACKAGE_ERLANG
  Depends on [n]: BR2_USE_MMU [=y] && !BR2_STATIC_LIBS [=n] && BR2_TOOLCHAIN_HAS_THREADS [=y] && BR2_PACKAGE_ERLANG_ARCH_SUPPORTS [=n]
  Selected by [y]:
  - BR2_PACKAGE_NERVES_CONFIG [=y]

WARNING: unmet direct dependencies detected for BR2_PACKAGE_ERLANG
  Depends on [n]: BR2_USE_MMU [=y] && !BR2_STATIC_LIBS [=n] && BR2_TOOLCHAIN_HAS_THREADS [=y] && BR2_PACKAGE_ERLANG_ARCH_SUPPORTS [=n]
  Selected by [y]:
  - BR2_PACKAGE_NERVES_CONFIG [=y]
#
# configuration written to /nerves/build/.config
#
make: Leaving directory '/nerves/env/platform/buildroot-2023.02.3'
------------

Build directory successfully created.

Configuration: /nerves/env/nerves_system_rpi0_video/nerves_defconfig

Next, do the following:
   1. cd /nerves/build
   2. make

For additional options, run 'make help' in the build directory.

IMPORTANT: If you update nerves_system_br, you should rerun this script.
nerves@b8054632fe06:/nerves/build$
```

I think this last part could be a bit more clear:

```
Configuration: /nerves/env/nerves_system_rpi0_video/nerves_defconfig

Next, do the following:
   1. cd /nerves/build
   2. make

For additional options, run 'make help' in the build directory.

IMPORTANT: If you update nerves_system_br, you should rerun this script.
```

The command already leaves you on `/nerves/build` so I'm not sure the cd command is needed.
Also I ran make and since the prompt looks different from OTP25 I wasn't sure if I could run make menuconfig like before (I tried and it's in fact the right place to run it).
Perhaps it could be nice to add a message indicating that here is where we run make menuconfig, etc?

guillego:
Also, not sure how much of a problem this is, but since the shell used to run within iex itself, exit it was "as intuitive" as pressing `ctrl+c`. I see now that we are in a bash prompt inside a docker so the way to exit is to type `exit`.
Perhaps there could be a message indicating this? Not sure if it's over doing it, but perhaps someone might appreciate it.

jjcarstens:
Ha, also funny you bring up `ctrl+c` - I actually always hated that. The typical idiom with Docker container I thought was to `ctrl+d` or `exit` so that `ctrl-c` can be used as expected in a terminal session

guillego:
No issue with the warnings! And I get the ctrl+c thing, but since it's the default way to exit iex, perhaps it would be good to add some comment.
In general I was referring to providing some extra context to the user who sees this for the first time since the experience is quite different from the one in OTP25, something like this:

```
warnings, info, etc
----

You are now in the nerves shell which is running inside Docker, to exit, type "exit" or press ctrl+d at any point.

Build directory successfully created.

Configuration: /nerves/env/nerves_system_rpi0_video/nerves_defconfig

Next, do the following: 
   1. cd /nerves/build (should be the current path already)
   2. make

For additional options, run 'make help' in the build directory.

Inside this shell is where you can run `make menuconfig` and related commands to configure your buildroot. Please refer to the docs (https://hexdocs.pm/nerves/customizing-systems.html#buildroot-package-configuration) for more information.

IMPORTANT: If you update nerves_system_br, you should rerun this script.
```

Alternatively we could just add more info in the actual docs to inform of these changes, let me know if you want a hand with those!

fhunleth:
Regarding:

```
WARNING: unmet direct dependencies detected for BR2_PACKAGE_ERLANG
  Depends on [n]: BR2_USE_MMU [=y] && !BR2_STATIC_LIBS [=n] && BR2_TOOLCHAIN_HAS_THREADS [=y] && BR2_PACKAGE_ERLANG_ARCH_SUPPORTS [=n]
  Selected by [y]:
  - BR2_PACKAGE_NERVES_CONFIG [=y]
```

You can ignore that. It's wrong and luckily doesn't actually do anything. It's due to building in an `aarch64` container and that's perfectly fine. It's just unexpected from Buildroot's point of view when compiling Erlang. I should be able to push a fix upstream if it hasn't already been addressed.

jjcarstens:
K, I've adjusted and added the prompts for the docker container. Turns out those warnings are to stderr, so we can leave `>/dev/null` and everyone wins!

image

I am hesitant to adjust the prompt for local linux builds. Before, it adjust for a transient shell. With this change, it would rewrite the whole active terminal (if fact, it may be desirable to have their configured one now show?)

guillego:
Looks perfect now! Basically the only difference is that you need to copy and paste the docker command (on Mac), so I think this is quite a good solution! Thanks a lot for working on this @jjcarstens

garazdawi:
Hi! Just saw your post over at [Elixir Forums](https://elixirforum.com/t/mix-nerves-system-shell-run-into-sh-1-exec-tty-sl-not-found/58190).

If you could open an issue on https://github.com/erlang/otp describing what you want to achieve, then we would be more than happy to take a look and see if we can help you. Either by figuring out the proper way to configure the new stdin/stdout system, or by adding something to Erlang that allows you to do what you want.

PR #902, opened 2023-08-29
Move `credo` and `dialyzer` to the `:lint` env

jjcarstens:
Dialyzer 1.4.0 dropped support for Elixir 1.11 - This allows us to still test Elixir 1.11 without failing to compile because of dialyxir which isn't used in those tests anyway

Also builds to shared directory so when developing or in CI it doesn't recompile the same stuff multiple times


fhunleth:
Turning off the build per environment worries me. It might be safe for this repository, but I can think of really easy ways to cause confusing errors and wouldn't want to it be copied around to other projects.

jjcarstens:
Ya, some of the other projects compile differently based on `dev` vs `test` (especially ones with a Makefile). This is really a minor optimization and maybe even mostly OCD knowing the tasks I use frequently will need its own compilation of the whole project (and sped more time in CI). I don't mind leaving it out, but thought I'd present the idea.

jjcarstens:
Went with a simpler approach. Maybe next time 😉

PR #904, opened 2023-09-04
Adjust compilation error when `nerves_bootstrap` missing

jjcarstens:
The same error was being used whether nerves_boostrap wasn't started or completely missing. This would lead to confusing error messages for users and unhelpful instructions in the case the archive is totally missing.

This changes that check to use a separate error if `nerves_bootstrap` is completely missing and needs to be installed

### DIFF
--- a/lib/mix/tasks/compile.nerves_package.ex
+++ b/lib/mix/tasks/compile.nerves_package.ex
@@ -17,8 +17,7 @@ defmodule Mix.Tasks.Compile.NervesPackage do
     if Nerves.Env.enabled?() do
       config = Mix.Project.config()
 
-      bootstrap_started?()
-      |> bootstrap_check()
+      bootstrap_check!()
 
       _ = Nerves.Env.ensure_loaded(Mix.Project.config()[:app])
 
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/compile.nerves_package.ex
```diff
@@ -41,31 +40,47 @@ defmodule Mix.Tasks.Compile.NervesPackage do
...
+              """
+            else
+              ""
+            end
+        )
+    end
+  end
 
-    Mix.raise(error)
+  defp boostrap_installed?() do
```
JasonTrue, line 79: Minor typo. s/boostrap/bootstrap/g

jjcarstens: Good catch. Thanks!

lib/mix/tasks/compile.nerves_package.ex
```diff
@@ -41,31 +40,50 @@ defmodule Mix.Tasks.Compile.NervesPackage do
...
-    error =
-      """
-      Compiling Nerves packages requires nerves_bootstrap to be started, which ought to
-      happen in your generated `config.exs`. Please ensure that MIX_TARGET is set in your environment.
-      """ <>
-        if in_umbrella?(File.cwd!()) do
+      not bootstrap_installed?() ->
+        Mix.raise("""
+        Compiling Nerves packages requires the nerves_bootstrap archive which is missing
+        from the Elixir version currently in use (#{System.version()}).
```
jjcarstens, line 51: @JasonTrue I altered the error message and added the Elixir version based on our Slack conv

fhunleth (approved):

PR #908, opened 2023-10-01
Update User Interfaces.md

Lucassifoni:
Some precisions on the relevant Phoenix configuration - the point isn't to give every possible variation on Phoenix configuration in the Nerves docs, but since the advice given is to copy some of Phoenix's required config in `target.exs`, this can result in confusion on some boards that do not have an easy way to give feedback before they fully boot. 

Maybe this is too much hand-holding ? The idea isn't to merge as-is but first discuss if that kind of "notice" has its place. In my case, I lost a few minutes before realizing that Swoosh defaulted to `hackney` on my board as its transport layer if you did not configure it to use Finch, and hackney isn't included in phoenix dependencies..

### DIFF
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -135,7 +135,7 @@ Note that this minimal configuration corresponds to our freshly generated
 phoenix application without Ecto or Swoosh as we passed the `--no-ecto` and
 `--no-mailer` flags to the generator earlier. If you wish to use those features, 
 remember to add the relevant configuration. An example for Ecto can be seen 
-in the [Nerves + Phoenix example]: https://github.com/nerves-project/nerves_examples/blob/1da4597bee5d9f26c643cb32523fc70e136d1e2b/hello_phoenix/firmware/config/target.exs#L17 .
+in the [Nerves + Phoenix example](https://github.com/nerves-project/nerves_examples/blob/1da4597bee5d9f26c643cb32523fc70e136d1e2b/hello_phoenix/firmware/config/target.exs#L17).
 
 There we have it! A Phoenix-based web application is now ready to run on our
 Nerves-based embedded device. By separating the Phoenix-based project from the
--- a/docs/User Interfaces.md
+++ b/docs/User Interfaces.md
@@ -44,7 +44,7 @@ mkdir my_app && cd my_app
 # Create a Nerves firmware project called "my_app_firmware"
 mix nerves.new my_app_firmware
 
-# Create a Phoenix 1.6 UI project called "my_app_ui"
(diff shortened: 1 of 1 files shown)


mnishiguchi (approved):
@Lucassifoni Thank you for your suggestion. I think it surely makes the explanation clearer, considering that some readers want full Phoenix settings.

@fhunleth @jjcarstens Maybe the link to the example can be more general for future proof? For example, it can be the config page of main branch, or README.md.

Lucassifoni:
In a more general way, what would be a proper place to post *specific* how-tos ? I'm running into problems linked to *specific* tasks that are sometimes asked on forums, like "how do I add support for X specific device driver", and solve them for my use case.

For those kind of problems, people asking questions could ask the more general and proper question, like *how do i add support for a device with buildroot*, but the specific one (like : how do I add support to Nerves for webcam X) is often posted. 

Maybe blog posts can fill this role, but I wonder if a guide that answers the general question with one or some specific examples would be useful ?

Those are just some thoughts as I fullfill my requirements step after step. I see other people on the internet running into the same problems months or years ago, so I'm wondering how to convert my specific resolution to more useful general how-tos..

mnishiguchi:
@Lucassifoni Sorry my comment might complicate things :sweat_smile: This PR is good as is. I was thinking about the link to the Ecto example source at first, but it seems fine.

> Maybe blog posts can fill this role, but I wonder if a guide that answers the general question with one or some specific examples would be useful ?

I think if you would like to share your Nerves tips with the communities, depending on the nature of the topic, it can be:
- a blog post (your own or maybe [Embedded Elixir](https://embedded-elixir.com))
- another simple example Nerve project that is dedicated to the topic in [nerves-project/nerves_examples](https://github.com/nerves-project/nerves_examples/tree/main)
- a small section in Nerves documentation with links to relevant resources

I agree that it would be nice if all the common tips were more visible to Nerves users somehow...

Lucassifoni:
> @Lucassifoni Sorry my comment might complicate things 😅 This PR is good as is. I was thinking about the link to the Ecto example source at first, but it seems fine.
> 
> > Maybe blog posts can fill this role, but I wonder if a guide that answers the general question with one or some specific examples would be useful ?
> 
> I think if you would like to share your Nerves tips with the communities, depending on the nature of the topic, it can be:
> 
>     * a blog post (your own or maybe [Embedded Elixir](https://embedded-elixir.com))
> 
>     * another simple example Nerve project that is dedicated to the topic in [nerves-project/nerves_examples](https://github.com/nerves-project/nerves_examples/tree/main)
> 
>     * a small section in Nerves documentation with links to relevant resources
> 
> 
> I agree that it would be nice if all the common tips were more visible to Nerves users somehow...

That's very clear and answers my question, thank you ! I'll try to summarize this project under a text form first then.

PR #913, opened 2023-10-12
Add support for Gitea release artifacts.

jimsynz:
This is mostly just a copy-pasta of the existing GitHub resolver with a few tweaks to work with the updated response format.

Notably, this PR introduces the idea of a "three slash repo", which allows you to specify the hostname of your instance, as well as the org and repo name.  For example: `gitea.com/jmshrtn/nerves_artifact_test`.

I have successfully tested downloading artifacts from a [repo on my Forgejo instance](https://code.harton.nz/bivouac/bivouac_system_x86_64).

Let me know if there's anything you'd like changed.


fhunleth:
Nice! Thank you for contributing Gitea support!

It makes sense to me based on my skim and me just learning about Gitea now. Only thing that I'd add is a mention in the docs at https://github.com/nerves-project/nerves/blob/main/docs/Customizing%20Systems.md#creating-an-artifact. It looks like it would be sufficient to just add it to the list of helpers. I'd assume someone familiar with Gitea would understand the 3-slash repo without any additional text, but feel free to add if I shouldn't assume. 

@jjcarstens may have comments since this is more his area.

lib/nerves/artifact/resolvers/gitea_api.ex
```diff
@@ -0,0 +1,138 @@
+defmodule Nerves.Artifact.Resolvers.GiteaApi do
```
jjcarstens, line 1:
```suggestion
defmodule Nerves.Artifact.Resolvers.GiteaAPI do
```

jjcarstens (approved):
Looks good to me! Just a small nit to change the module name to `GiteaAPI` to match the other modules casing as well

jimsynz:
Sorry for the radio silence. I had to travel for work last week. I'll make the changes suggested then we should be g2g.

jjcarstens (approved):
🎉 Thanks!

PR #916, opened 2023-10-18
Update Python version in Installation.md

jshprentz:
Ubuntu 22.04.3 LTS cannot install package `python`. Install package `python3` instead. Add a section heading instead of continuing the fwup section. Use consistent capitalization for "Nerves system" and "Nerves systems."


fhunleth:
Thanks for the update! Can't believe we still had non-`python3` reference.

PR #922, opened 2023-10-31
The Elixir community moved from Slack to Discord

joshk:
A small docs update which makes it easier to find the real hiding place of the Nerves, and Elixir, community.


fhunleth:
Hmmmmm.... Give me a chance to do some asking around. I see quite a bit more activity on Slack than either Discord or the Forum. However, you're right, elixir-lang.org no longer points to Slack and joining the elixir-lang slack doesn't seem easy anymore.

fhunleth:
I found out that the reason for elixir-lang.org removing Slack was due to the signup page not working. See https://github.com/elixir-lang/elixir-lang.github.com/commit/3a4be9a8626eb5654e54b2bab3cfe7b3d5838004. That was in August and I couldn't find any progress on the situation. It seems that people are giving out invite links now. See https://elixirforum.com/t/unable-to-join-the-slack-channel/56331/11. 

I think your PR is the right thing to do, and if the situation ever changes, we can revisit it then.

PR #925, opened 2023-11-02
Rename docs to guides, and group similar guides

joshk:
I totally understand if this PR is going too far and isn't seen as a good addition.

I took some learnings from how Livebook and Liveview structure their extra pages and applied them to the Nerves project.

This PR:
- uses the naming of Guides in the top of the sidebar
- moves the pages from `docs` to `guides` (in the repo)
- adds group sections of `Introduction`, `Core`, and `Advanced`
- simplified a bunch of titles in various guides so they display better in the sidebar
- removed the 'Is something wrong?` section at the bottom of all guides, these weren't formatted correctly and seemed redundant. 
- updated all guide filenames to not use spaces

This is a screenshot of what the sidebar now looks like:

Screenshot 2023-11-01 at 5 36 56 PM


mnishiguchi (approved):
Wow I like the way the sidebar looks after your changes! 😍 Also I think the introduction of the categories will enable us to further improve the organization of our documents in the long run.

The term "Core" sounds kind of unclear to me but I guess we can think about it anytime.

Thank you for your great work 👍

joshk:
Thanks for your feedback @mnishiguchi :)

I also struggled with `Core` but thought it would be a good first change.

joshk:
(just forced pushed a little change to my PR where I noticed that I had `Examples projects` instead of `Example projects`)

joshk:
Hi @jjcarstens, do you have any feedback on this PR?

jjcarstens:
Sorry I'm so delayed. I'm looking at this now and may have a few thoughts shortly

jjcarstens (approved):
Okay, I've stared and thought a bit about this and I think maybe `Usage` makes more sense than `Core`, but I'm good as-is to get rolling.

I really like our directory structure in the [Nerves README.md](https://github.com/nerves-project/nerves?tab=readme-ov-file#nerves-projects) (I think we'll eventually want that in the docs maybe?). I was trying to think how to incorporate some of that and totally went wild with a new outline/guide structure that I put in https://github.com/nerves-project/nerves/issues/926 to maybe adjust further in future PR's without holding this up

fhunleth:
This is looking great! Thank you.

joshk:
Thanks for the feedback @jjcarstens, I can create another PR for the `Usage` change, and to incorporate some of the ideas in #926 

Thanks for approving these changes and merging in the PR.

PR #929, opened 2023-11-11
Add mix task to load nerves_defconfig

amclain:
This PR adds the command `mix nerves.loaddefconfig` to copy changes made to `nerves_defconfig` over to the Nerves artifact. It does the opposite of running `make savedefconfig` from the artifact.

### DIFF
--- a/lib/mix/tasks/nerves.loaddefconfig.ex
+++ b/lib/mix/tasks/nerves.loaddefconfig.ex
@@ -1,0 +1,33 @@
+defmodule Mix.Tasks.Nerves.Loaddefconfig do
+  @shortdoc "Load nerves_defconfig into the Nerves artifact"
+
+  @moduledoc """
+  Load the nerves_defconfig configuration into the Nerves artifact.
+  This is the opposite of `make savedefconfig`.
+
+  ## Usage
+
+      mix nerves.loaddefconfig
+
+  """
+  use Mix.Task
+  import Mix.Nerves.IO
+
+  @impl Mix.Task
+  def run(_argv) do
+    debug_info("loaddefconfig Start")
+    Nerves.Env.disable()
+
+    pkg_name = Mix.Project.config()[:app]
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/nerves.loaddefconfig.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Mix.Tasks.Nerves.Loaddefconfig do
+  @shortdoc "Load nerves_defconfig into the Nerves artifact"
+
+  @moduledoc """
+  Load the nerves_defconfig configuration into the Nerves artifact.
+  This is the opposite of `make savedefconfig`.
```
jjcarstens, line 6:
```suggestion
  @shortdoc "Load nerves_defconfig into the Nerves artifact"

  @moduledoc """
  Load the nerves_defconfig configuration into the Nerves artifact.
  This is the opposite of `make savedefconfig`.
```

I don't think this is quite right. This is only used during build and development but the phrasing of this doc makes it sound like I can load into an already compiled artifact which is not correct.

Maybe mirroring as `nerves.create_build` would be better? Or just changing doc to reflect this loads configuration to prepare for artifact build?

jjcarstens (changes requested):

fhunleth:
I feel this is an advanced feature since you have to know what you're doing since not all Buildroot configuration updates will actually rebuild what you've changed. In fact, I field questions on it with some regularity, so making a helper to encourage this workflow feels like it will create more support requests. The better answer is to erase and start over.

Since we talk about removing in tree Buildroot development on an annual basis, perhaps this is a better addition to `nerves_systems`. And then add an explicit warning that this is a best effort feature and if it doesn't work, then do a full rebuild.

amclain:
Sure, having `nerves_systems` do this instead makes sense. I'll move it over there and update it to work with the out of tree builds.

PR #932, opened 2023-11-15
Fix potential error in Nerves and Phoenix odcs

mvkvc:
I followed the guide to set up Nerves and Phoenix together and kept getting an error on this line `{:phoenix_live_reload, "~> 1.2", only: :dev && Mix.target() == :host},`. There was a compiler warning and also the changed way is how it is in `nerves_examples` [here](https://github.com/nerves-project/nerves_examples/blob/6820ac37c3d5f77cf83be12eba63c085961f1390/hello_phoenix/ui/mix.exs#L41C1-L41C52). Could be wrong though so happy to hear if so, either case it works for me now with the change.

### DIFF
--- a/guides/core/user-interfaces.md
+++ b/guides/core/user-interfaces.md
@@ -84,7 +84,7 @@ host:
     [
       {:phoenix, "~> 1.6.0"},
       # ...
-      {:phoenix_live_reload, "~> 1.2", only: :dev && Mix.target() == :host},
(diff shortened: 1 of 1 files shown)


jjcarstens:
I think we need to make it match the ESbuild and tailwind dependencies. This brings in an executable that should only run on host. Including it on device is usually what breaks things and leaving as `only: :dev` seems like it leaves that potential possibility

PR #944, opened 2024-01-08
boolean logic edit

bradhanks:
ran mix test before and after


fhunleth:
This change feels like it's personal preference and doesn't fix anything. If this came along with a more substantial change with the same code, then I'd probably merge it. However, in an abundance of caution, I'd like to leave the code as is for now to avoid even thinking about ramifications.

Issue #876, opened 2023-06-02
`mix firmware` failure during squashfs


jjcarstens:
With squashfs 4.6+ on the host machine, some `mix firmware` builds may fail after building the rootfs and then squashing it

```
|nerves_bootstrap| Building OTP Release...

* skipping runtime configuration (config/runtime.exs not found)
* creating _build/rpi0_dev/rel/my_fw/releases/0.1.0/vm.args
Updating base firmware image with Erlang release...
Copying rootfs_overlay: .../_build/rpi0_dev/nerves/rootfs_overlay
Copying rootfs_overlay: .../my_fw/rootfs_overlay
** (Mix) Nerves encountered an error. %IO.Stream{device: :standard_io, line_or_bytes: :line, raw: true}
```

The workaround is to downgrade squashfs to 4.5 and things _should_ work again.

We should investigate the breakage with squashfs 4.6 and how to mitigate

LostKobrakai:
The bit of debugging I did seemed to fail in a `unsquashfs` call of `~/.nerves/artifacts/our-fw/scripts/merge-squashfs`

jjcarstens:
I was able to reproduce this with a private repo. Once digging through the scripts and commands being run during squashing, this is what I ran into

```
FATAL ERROR: write_file: file squashfs-root/usr/lib/xtables/libip6t_hl.so already exists
```

Frank gave me the idea to check for files without case-sensitivity and sure enough, there are 2

```
> my_fw √ % grep -i libip6t_hl /var/folders/l9/61kntssd7g19yw5h4l9qxsr00000gn/T/tmp.P40vXDGq/pseudofile
/usr/lib/xtables/libip6t_HL.so m 755 root wheel
/usr/lib/xtables/libip6t_hl.so m 755 root wheel
```

MacOS is case-insensitive by default. This actually hits at a lots of random times. For cases during squashfs, it's probably running into files like this that are provided by the underlying Nerves system and is going to vary widely.

I need to think about how to handle this. I'm assuming the recent squashfs update has some sort of change to ignore case and just need to see how to work around it

schrockwell:
For folks who need to downgrade squashfs on macOS using Homebrew:

```sh
brew tap-new zappy/edith
brew extract --version="4.5.1" squashfs zappy/edith
brew install squashfs@4.5.1
brew link --overwrite squashfs@4.5.1
unsquashfs -version
```

SteffenDE:
[...]

PR #951, opened 2024-02-07
Add docs for building systems on macOS, and $TMPDIR

schrockwell:
This is a follow-up to #876 and https://github.com/nerves-project/nerves_system_br/pull/778 that adds documentation specifically about building custom systems on macOS, and the new `$TMPDIR` environment variable.

Should the docs recommend setting both `$NERVES_DL_DIR` and `$NERVES_ARTIFACTS_DIR`, or should we just set `$XDG_DATA_HOME`?

### DIFF
--- a/guides/advanced/customizing-systems.md
+++ b/guides/advanced/customizing-systems.md
@@ -420,3 +420,15 @@ You can also use the GitHub interface to do this:
 ```text
 https://github.com/YourGitHubUserName/custom_rpi3/compare/main...nerves-project:main?expand=1
 ```
+
+## Building Systems on macOS
+
+The primary Apple File System (APFS) volume on Macs is case-insensitive, but the Nerves filesystem is case-sensitive. This can cause filename conflicts when including some packages.
+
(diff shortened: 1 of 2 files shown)


schrockwell:
Yup, I think that all makes sense. Thanks.

jjcarstens:
Why do the download and artifact directories need to be on the volume? I thought all that might be needed is `TMPDIR` since the build and `mktemp` would be using the copied temp files which would be available in case sensitive?

I don't think `NERVES_DL_DIR` needs to be recommended since that is just compressed tar and zip files.  `NERVES_ARTIFACTS_DIR` is fine but we could also say `XDG_DATA_HOME` would be recommended since that would handle all the Nerves pieces

schrockwell:
@jjcarstens You may be right about downloads, but for artifacts I did find conflicts such as

```
/Volumes/Nerves/artifacts/nerves_system_rpi3a_tailscale-portable-1.25.1/staging/usr/lib/xtables/libip6t_hl.so
/Volumes/Nerves/artifacts/nerves_system_rpi3a_tailscale-portable-1.25.1/staging/usr/lib/xtables/libip6t_HL.so
```

I'm happy to change it to `XDG_DATA_HOME` to make it simpler and potentially more future-proof.

PR #952, opened 2024-02-08
`only: env` doesn't accept booleans just the environment use instead `targets`

tyoc213:


fhunleth:
Thanks for the PR. We definitely missed catching that line when copying from the examples to the docs. I just looked and the examples are the same as what you wrote.

PR #954, opened 2024-02-09
update compatibility chart

mnishiguchi:
### Description

- adjust the script for the changes we made in guides recently (file path change)
- update the Compatibility chart

### Quick test

The content is correct; latest bbb nerves system is [2.20.2](https://hex.pm/packages/nerves_system_bbb/2.20.2) as of this writing.

CleanShot 2024-02-09 at 15 03 03


fhunleth:
Thanks!

PR #955, opened 2024-02-10
`only: env` doesn't accept booleans use instead `targets`

tyoc213:


fhunleth:
@tyoc213 I'm confused. This looks the same as #952 which has been merged.

tyoc213:
@fhunleth indeed!!

Issue #962, opened 2024-03-25
Following the User Interfaces Tutorial for Phoenix Does Not Work


panda2k:
### Environment

* Elixir version (`elixir -v`): `1.16.1`
* Nerves environment: (`mix nerves.env --info`): `MIX_TARGET: rpi3`, `MIX_ENV: dev`

### Current behavior
When following the tutorial line by line, running `mix deps.get` in `my_app_firmware` results in the following error: 
```
Dependencies have diverged:
* phoenix (Hex package)
  the :targets option for dependency phoenix

  > In /Users/michael/Programming/my_app/my_app_ui/mix.exs:
    {:phoenix, "~> 1.7.11", [targets: [:rpi, :rpi0, :rpi2, :rpi3, :rpi3a, :rpi4, :bbb, :osd32mp1, :x86_64, :grisp2, :mangopi_mq_pro], env: :prod, hex: "phoenix", repo: "hexpm"]}

  does not match the :targets option calculated for

  > In deps/phoenix_live_reload/mix.exs:
    {:phoenix, "~> 1.4", [only: :dev, targets: [:host], env: :prod, hex: "phoenix", repo: "hexpm", optional: false]}

  Ensure you specify at least the same targets in :targets in your dep
** (Mix) Can't continue due to errors on dependencies
```

This can be fixed by doing  `{:my_app_ui, path: "../my_app_ui", targets: [:host | @all_targets], env: Mix.env()},` in `my_app_firmware/mix.exs` (note `[:host | @all_targets]` instead of `targets: @all_targets`). 

However, after doing this `RingLogger.next` reveals a new error when starting the server: 
```
MyAppUi.Application.start(:normal, []) returned an error: shutdown: failed to start child: MyAppUiWeb.Endpoint
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function Plug.Cowboy.child_spec/1 is undefined (module Plug.Cowboy is not available)
            Plug.Cowboy.child_spec([ref: MyAppUiWeb.Endpoint.HTTP, scheme: :http, plug: {MyAppUiWeb.Endpoint, []}, options: [port: 80, otp_app: :my_app_ui]])
            (phoenix 1.7.11) lib/phoenix/endpoint/cowboy2_adapter.ex:92: Phoenix.Endpoint.Cowboy2Adapter.child_spec/4
[...]

PR #963, opened 2024-03-25
Update User Interface Guide to Use Latest Version of Phoenix

panda2k:
Solves #962 

This PR updates the `user-interfaces.md` guide to use the latest version of Phoenix (1.7), which also uses `bandit` instead of `plug_cowboy` for the HTTP server.

### DIFF
--- a/guides/core/user-interfaces.md
+++ b/guides/core/user-interfaces.md
@@ -61,7 +61,7 @@ as a deployment wrapper around the `my_app_ui` project.
       # Dependencies for all targets
       {:nerves, "~> 1.7.0", runtime: false},
       # ...
-      {:my_app_ui, path: "../my_app_ui", targets: [:host | @all_targets], env: Mix.env()},
+      {:my_app_ui, path: "../my_app_ui", env: Mix.env()},
       # ...
     ]
   end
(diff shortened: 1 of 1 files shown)


guides/core/user-interfaces.md
```diff
@@ -61,7 +61,7 @@ as a deployment wrapper around the `my_app_ui` project.
       # Dependencies for all targets
       {:nerves, "~> 1.7.0", runtime: false},
       # ...
-      {:my_app_ui, path: "../my_app_ui", targets: @all_targets, env: Mix.env()},
+      {:my_app_ui, path: "../my_app_ui", targets: [:host | @all_targets], env: Mix.env()},
```
fhunleth, line 64:
```suggestion
      {:my_app_ui, path: "../my_app_ui", env: Mix.env()},
```
It seems like the `:targets` field could be dropped if `:my_app_ui` is going to be pulled in for everything. @mnishiguchi or @jjcarstens - do either of you remember why `:targets` was added in the first place?

panda2k: I just tested this and it works fine with `:targets` field dropped and it worked fine. Unless there is a reason to specify it explicitly, I agree it should be dropped.

mnishiguchi (approved):

guides/core/user-interfaces.md
mnishiguchi: @fhunleth I verified that [our official phoenix example](https://github.com/nerves-project/nerves_examples/blob/8cc860beefe2e389e1dccd1efb4ef8f43b1a794d/hello_phoenix/firmware/mix.exs#L56) does not have the `:target` field there.

```elixir
      {:ui, path: "../ui", env: Mix.env()},
```

PR #972, opened 2024-05-12
docker: Don't use ssh-agent, instead mount ~/.ssh

ringlej:
Building on a mac will result in this failure:
```
docker: Error response from daemon: error while creating mount source path '/host_mnt/private/tmp/com.apple.launchd.lINaqsIhwG/Listeners': mkdir /host_mnt/private/tmp/com.apple.launchd.lINaqsIhwG/Listeners: operation not supported.

```
This is failing because Apple doesn't support passing a unix socket across a hypervisor. This happens when trying to use SSH_AUTH_SOCK to mount the ssh-agent (unix) socket for authentication.

Instead, mount the user's ~/.ssh directory so that ssh has access to the user's private keys during the docker container instance running the build.


jjcarstens (approved):
LGTM 🎉

PR #978, opened 2024-05-27
Build from from `rootfs.tar` system files

jjcarstens:
Starts support to allow building firmware from a `rootfs.tar` tarball file more efficiently in Elixir, rather than relying on bash scripts to unsquash, combine, and resquash an FS. 

The `mix firmware` task was due for a good refactoring, so it looks like a lot. The old way of building is maintained for the typical `rootfs.squashfs` file that systems have traditionally provided.

It also adjusts the build process to allow using EROFS instead of squashfs when making the firmware

> [!NOTE]
> This adds `config :nerves, :firmware, fs_type: :erofs` to specify to use EROFS. If there is a preferred option name and/or location, let me know
>
> Other pieces to still consider:
> - [ ] Scrubbing the OTP release (replacing [`scrub-otp-release.sh`](https://github.com/nerves-project/nerves_system_br/blob/main/scripts/scrub-otp-release.sh))
> - [ ] If priorities need to be adjusted or [these Erlang pieces added](https://github.com/nerves-project/nerves_system_br/blob/main/scripts/rel2fw.sh#L45-L50) 
> - [ ] Maybe support building `*.img` firmware file [like the end of `rel2fw.sh`](https://github.com/nerves-project/nerves_system_br/blob/main/scripts/rel2fw.sh#L165-L179)
> - [ ] Keeping the same `echo` statements from [`rel2fw.sh`](https://github.com/nerves-project/nerves_system_br/blob/main/scripts/rel2fw.sh)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
             - run: mix format --check-formatted
             - run: mix deps.unlock --check-unused
             - run: mix docs
-            - run: mix hex.build
+            # - run: mix hex.build
             - run: mix test || mix test
             - run: mix credo -a --strict
             - run: mix dialyzer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - run:
           name: Install system dependencies
-          command: apk add --no-cache build-base procps
+          command: apk add --no-cache build-base procps git
       - checkout
       - run:
           name: Install hex, rebar, and nerves_bootstrap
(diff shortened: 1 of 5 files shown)


fhunleth:
Cool! Thank you. I believe tar_merger is still private so I'll copy over and rename. It has a number of deficiencies that I forgot to tell you about but it's getting close and this will definitely help me.

jjcarstens:
I added a few more notes. This could probably be split into 2 PRs for the EROFS concern, but after the redact that piece is so small that it could really go either way ¯\_(ツ)_/¯

I've tested manually, but probably worth adding some tests if I can think of a good way how

fhunleth:
I made a pass at rebasing this PR in the firmware-build-from-tar2 branch.

fhunleth:
Closing since code is in the v2 branch.

PR #985, opened 2024-06-24
Update compatibility chart

mnishiguchi:
Update the compatibility chart by running the script `./scripts/nerves_system_compatibility.exs`.

```
cd path/to/nerves
mix deps.get
./scripts/nerves_system_compatibility.exs --format html --doc
```

## Testing

Compare the contents before and after running the script.

```
mix docs --open
```

```
open https://hexdocs.pm/nerves/systems.html#compatibility
```

Here is the result. Looking good to me.

CleanShot 2024-06-24 at 20 05 19@2x


fhunleth:
Thanks!

PR #992, opened 2024-07-07
support setting the product in `mix.exs` using `:product`

joshk:
favor using `:product` in `mix.exs` for setting the product variable in the metadata


joshk:
a related question, is there a way to set the `product` metadata via an env var or `mix firmware` arg?

fhunleth:
This seems like a good change. I know I've wanted to decouple the app name and the product name in the past for aesthetic reasons. There's no way of override the product metadata via an environment variable. It should be possible to make `NERVES_FW_PRODUCT` at build time if wherever that's set has its semantics change to "set only if unset".

As for this PR, I assume there's at least one place in the docs that needs to get updated.

joshk:
After talking with Frank, we have decided on a different path forward when it comes to firmware verification on NervesHub, which this PR was related to.

PR #998, opened 2024-07-29
External Buildroot Packages

amclain:
Related to: nerves-project/nerves_system_br#806

This PR allows a Nerves system maintainer to add external Buildroot packages to a Nerves system so that Buildroot can be extended without forking [nerves_system_br](https://github.com/nerves-project/nerves_system_br). External Buildroot customizations are explained in [section 9.2](https://buildroot.org/downloads/manual/manual.html#outside-br-custom) of the Buildroot manual. Since Nerves sets `BR2_EXTERNAL` and controls the build process, this PR discovers Elixir dependencies with the project setting `buildroot_package: true` and adds them to the `BR2_EXTERNAL` variable.


fhunleth:
Thanks for this PR. It's clear that some documentation is missing. You can already add Buildroot packages without forking `nerves_system_br`, though. See https://github.com/nerves-project/nerves_system_bbb/tree/main/package/extra-dts for a really simple example, but hopefully it gives an idea about how you can add to the external menu.

amclain:
@fhunleth that example is within the BBB Nerves system though, right? We're looking to add a library that can be included across multiple Nerves systems that adds to the Buildroot menu. A good example is like what [nerves-config](https://github.com/nerves-project/nerves_system_br/tree/main/package/nerves-config) does, if it were maintained outside of the Nerves tooling.

fhunleth:
Let's move this to #nerves-dev. I think there's a lot to discuss on the topic and I'm not sure if the fidelity of GitHub comments will make it easy.

amclain:
As discussed, we'll keep the new menus in the Nerves systems for now. Thanks!

PR #1002, opened 2024-08-18
Updated links for NervesSystemRPi5

cocoa-xu:
Was looking for Nerves system for SiFive and generic ARMv7 and ARM64 devices and noticed this one, pretty sure NervesSystemRPi5 is available now ;)


fhunleth:
Thanks

PR #1007, opened 2024-09-17
Removes reference to the discord channel, channels users to the Nerves subforum on Elixirforum.com

Lucassifoni:


fhunleth:
Good catch. Thank you for updating this!!!

Lucassifoni:
Hi @fhunleth , I see that there are tons of other references to the Discord channel. I'll send PRs for each one to make it clearer to newcomers !

PR #1009, opened 2024-09-17
Add reference to mise; update Elixir/Erlang install versions

fhunleth:


jjcarstens:
Just pushed a commit rather than trying to make the suggestion UI work. It puts the asdf and mine instructions into tabs instead and moves some of the notes outside of the shell block into tip admonitions

fhunleth:
@jjcarstens That's neat. Thanks for adding that.

PR #1011, opened 2024-09-26
Force `TERM` for docker system shell

### LINKED ISSUE #1004: Nerves Shell Broken
### Environment

Elixir 17.0.2
Erlang 27.0.1
macOS 14.2.1

### Current behavior

The Nerves Shell Docker container runs but I am unable to configure the system with `make menuconfig` or `make linux-menuconfig`

`make menuconfig` seems to partially load the Buildroot configuration tool but I only see a line of text along the bottom
`make linux-menuconfig` seems to start building linux rather then load the configuration tool

### Expected behavior

Should have the same behavior as before where the system can be configured

jjcarstens:
Fixes #1004

EDIT:
Turns out we [added `TERM=dumb` to the `nerves_system_br` image starting with v1.28](https://github.com/nerves-project/nerves_system_br/commit/db6f9a2a5848ca75f26c9c4a3b2a7bf80e606da1#diff-181ba968e8b37dc7c64319846e544967221e3b5c1ec96bd070456abe37b2ccf2R17) based on CircleCI Docker template recommendations. This is nice to have for CI envs which is the majority of uses for that image, so this adjusts the Nerves tooling to put the `TERM` value of the user in _or_ default to `xterm-256color` if not set

### DIFF
--- a/lib/nerves/artifact/build_runners/docker.ex
+++ b/lib/nerves/artifact/build_runners/docker.ex
@@ -312,7 +312,13 @@ defmodule Nerves.Artifact.BuildRunners.Docker do
   defp env(:root), do: env(0, 0)
 
   defp env(uid, gid) do
-    ["--env", "UID=#{uid}", "--env", "GID=#{gid}"]
+    term =
+      case System.get_env("TERM") do
+        term when is_binary(term) and byte_size(term) > 0 -> term
+        _ -> "xterm-256color"
+      end
+
+    ["--env", "UID=#{uid}", "--env", "GID=#{gid}", "--env", "TERM=#{term}"]
   end
 
   defp uid() do


fhunleth:
It seems like we should check the users environment to see if `TERM` is set and then pass that. Given the confusion of not having `TERM` set, this seems like a reasonable default for most of our users. 

I assume the Docker stopped automatically passing the user's environment through at some point given that this used to work and now it doesn't. Either that or `TERM` was selectively passed as a convenience and now they made it explicit?

jjcarstens:
You're right. Digging deeper, this stopped working in v1.28.0 of nerves_system_br. When I use the nerves_system_br v1.27.3 image

```sh
> nerves_system_rpi4 √ % docker run -ti --rm ghcr.io/nerves-project/nerves_system_br:1.27.3 /bin/bash
root@8b657537db29:/# echo $TERM
xterm
```

And using v1.28.0 image

```sh
> nerves_system_rpi4 √ % docker run -ti --rm ghcr.io/nerves-project/nerves_system_br:1.28.0
nerves@4a50e7a29ec8:~/project$ echo $TERM
dumb
```

This is the [Dockerfile diff between those versions](https://github.com/nerves-project/nerves_system_br/compare/v1.27.3...v1.28.0#diff-181ba968e8b37dc7c64319846e544967221e3b5c1ec96bd070456abe37b2ccf2) - Haven't figured out specifically what caused the change

PR #1015, opened 2024-10-04
Update User Interfaces to reference phoenix examples

lawik:


lawik:
Said I'd update this to reference hello_liveview and clean out poncho and umbrella stuff that we might not want to confuse people.

fhunleth (approved):

PR #1016, opened 2024-10-11
Added info on how to use lxqt-openssh-lastpass in shell.nix.

eamonno:
Added specifics on how to swap to using lxqt-openssh-lastpass to the NixOS section.

Also fixed one small typo.


fhunleth (approved):
Thanks!

PR #1018, opened 2024-10-21
Refresh Linux install and install fwup via asdf

fhunleth:
This moves the Linux distro-specific parts to tabs to reduce the amount
of text the user sees. It installs a little more in the first package
installation line to cover packages the user may need for building
Erlang and fwup as well.

The big change for Debian and Fedora users is to use asdf to install
fwup which should be easier than the current build-from-sourse
instructions.

Here's what it looks like:

image

### DIFF
--- a/guides/introduction/installation.md
+++ b/guides/introduction/installation.md
@@ -63,32 +63,58 @@ choco install fwup /y
 
 When running on WSL2, Nerves uses the Linux version of `fwup` for building the
 firmware files and the Windows version of `fwup` for burning firmware to SD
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):
This looks good to me. Though I could also see fwup being its own section with tabs for asdf, mise, or compiling from source rather than jamming it into each OS tab. Though I don't feel strongly about that

fhunleth:
Arch and NixOS have fwup packages. Alpine does too, but we didn't have a section for it. I think that asdf-fwup will end up being used in more places, but I view it as a workaround for Debian and Ubuntu.

PR #1019, opened 2024-10-21
Add mix discovery command for device discovery

lawik:
Uses :mndp library and delegates to mix mndp.discover. If we switch discovery method over time this means we can fairly easily switch without changing the command.

It requires MNDP to be running in your app and a matching change to mix nerves.new might be reasonable to include mndp.

### DIFF
--- a/lib/mix/tasks/discover.ex
+++ b/lib/mix/tasks/discover.ex
@@ -1,0 +1,28 @@
+defmodule Mix.Tasks.Discover do
+  @moduledoc """
+  Find Nerves devices on the local network.
+
+  Should find across wired network, wireless and even USB gadget mode.
+
+  Currently using the MNDP library implementing the MicroTik Neighbor
+  Discovery Protocol. This is more reliable than mDNS which often gets
+  tangled up with general usage by the system.
+  """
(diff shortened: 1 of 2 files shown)


lawik:
I am not sure we are happy to put this in yet or if mndp needs some time to bake. I really like what it would bring in terms of dev convenience.

You run `mix discover` and get:

Screenshot_20241021_103654

This doesn't have the risk of collision with system implementations and general weirdness that we get with mDNS. Maybe we will find other weirdness? I don't know. But it seems rad.

lawik:
nerves_bootstrap PR: https://github.com/nerves-project/nerves_bootstrap/pull/332

fhunleth:
@lawik I really like MNDP, but I think I need to spend some time using this in anger before elevating it to Nerves core tooling. The good part is that everyone can use `mix mndp.discover` right now without any changes to Nerves.

lawik:
Entirely fair. I'm fine with these PRs idling or being closed and we go in for it when it feels ready.

fhunleth:
We're moving to use igniter for installing everything, so closing in anticipation for working on that.

PR #1030, opened 2024-11-24
Fix link to Hello LiveView example

clupprich:
I noticed that the link to the Hello LiveView example is broken. This change fixes that.


fhunleth (approved):
Thanks!

PR #1032, opened 2024-12-04
Docs/nerves systems guide improvement

Damirados:
This is still WIP.

Started to take it into direction where it completely replaces  "Customizing your nerves system guide"

### DIFF
--- a/guides/advanced/building-sytems.md
+++ b/guides/advanced/building-sytems.md
@@ -388,8 +388,67 @@ If your customizations are for long-term use, consider committing your changes t
 - Reproducing builds in the future.
 
 Example:
+Let's say that you want to version control your customized rpi3 system
 
 ```bash
-git add config/nerves_defconfig
-git commit -m "Customize Buildroot configuration for <system name>"
+cd src
+cp -r nerves_system_rpi3 custom_rpi3
 ```
+
+The name of the system directory is up to you, but we will call it `custom_rpi3`
+in this example. It's recommended that you check your custom system into your
+version control system before making changes. This makes it easier to merge in
+upstream changes from the official systems later. For example, assuming you're
+using GitHub:
+
+
+```bash
+# After creating an empty custom_rpi3 repository in your GitHub account
+
+cd custom_rpi3
+git remote rename origin upstream
+git remote add origin git@github.com:YourGitHubUserName/custom_rpi3.git
+git checkout -b main
+git push origin main
+```
+
+
+Next, tweak the metadata of your Mix project by updating your `mix.exs` with the following:
+
+* The module name of the mix project at the top of the file
+* the value of `@app` to `custom_rpi3`
+* the value of `@github_organization` to your GitHub user name or organization
+
+See the [Official Mix.Project](https://hexdocs.pm/mix/Mix.Project.html) document
+for the structure of this file.
+
+```elixir
+# custom_rpi3/mix.exs
+
+# defmodule NervesSystemRpi3.MixProject do
+defmodule CustomRpi3.MixProject do
+  #      =^^^^^^^^^^= Rename `NervesSystemRpi3` to `CustomRpi3`
+  use Mix.Project
+
+  # @github_organization "nerves-project"
+  @github_organization "YourGitHubUserOrOrganizationName"
+  #                    =^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^=
+  #                    Rename `"nerves-project"` here to your GitHub user or organization name
+
+  # @app :nerves_system_rpi3
+  @app :custom_rpi3
+  #    =^^^^^^^^^^^= Rename `nerves_system_rpi3` here to `custom_rpi3`
+end
+
+# =^^^= The rest of this file remains the same
+```
+
+```bash
+# Commit and push your changes.
+
+git add mix.exs
+git commit -m "Change project info"
+git push
+```
+
+Now you can go to your `nerves_systems/config/config.exs` and add it to your systems.
--- a/guides/advanced/building-sytems.md
+++ b/guides/advanced/building-sytems.md
@@ -25,17 +25,29 @@ The `nerves_systems` build process only works on **Linux** systems with `x86_64`
 
 Install the following packages in your Linux environment:
 
+#### Apt based distributions (Ubuntu, Debian...)
 ```bash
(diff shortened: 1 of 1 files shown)


guides/advanced/building-sytems.md
```diff
@@ -277,34 +289,70 @@ Customizing your Nerves system is an advanced but powerful way to tailor the sys
...
+    usually something like `$(NERVES_DEFCONFIG_DIR)/linux-x.y_defconfig`.
+
+3. After `make busybox-menuconfig`:
+
+    Unfortunately, there's not currently an easy way to save a BusyBox defconfig.
+    What you have to do instead is save the full BusyBox config and configure it
+    to be included in your `nerves_defconfig`.
+
+    ```bash
+    cp build/busybox-1.27.2/.config ../src/<full system name>/busybox_defconfig
```
Damirados, line 337: Using `<full system name>` here is not ideal.

Likely scenario is that you would want to add new system to `config.exs`
run
```bash
mix ns.clone
# Then actually just populate o directory with something like
mix ns.build --configure
# where it would just create o directory with config without building it
```

Ideally that would provide us with shell script that could export `NERVES_DEFCONFIG_DIR`
then this could be.
```bash
cp build/busybox-1.27.2/.config ${NERVES_DEFCONFIG_DIR}/busybox_defconfig
```

guides/advanced/building-sytems.md
```diff
@@ -341,8 +388,67 @@ If your customizations are for long-term use, consider committing your changes t
...
+
+```bash
+# Commit and push your changes.
+
+git add mix.exs
+git commit -m "Change project info"
+git push
+```
+
+Now you can go to your `nerves_systems/config/config.exs` and add it to your systems.
```
Damirados, line 454: This should include another section of how to now use that add custom system in a project and use it.

guides/advanced/building-sytems.md
Damirados: It would also be good idea for build to accept `--only`
```bash
mix ns.build --only custom_rpi3
```

guides/advanced/building-sytems.md
fhunleth: Good idea.

guides/advanced/building-sytems.md
Damirados: So I tried this with rpi3 system and
```bash
make busybox-update-config
```
results in
```
cp -f /home/damir/Projects/nerves/nerves_systems/o/rpi3/build/busybox-1.36.1/.config /home/damir/Projects/nerves/nerves_systems/src/nerves_system_br/board/nerves-common/busybox.config
```

Which now applies it to all other systems which is IMO not wanted behavior.

In case when fragments are set it does error
```
Unable to perform busybox-update-config when fragment files are set
make[1]: *** [package/busybox/busybox.mk:464: busybox-update-config] Error 1
make: *** [Makefile:23: _all] Error 2
```

How this guide states it in current version would be correct for using fragments.

I guess my question is what would be correct way to automate
```bash
cp build/busybox-1.27.2/.config ../src/<full system name>/busybox_defconfig
```

Maybe add new make option `make busybox-update-fragment-config that would basically call
```bash
cp build/busybox-1.27.2/.config ${NERVES_DEFCONFIG_DIR}/busybox_defconfig
```
and maybe set fragments at the same time.

fhunleth: Oh, you're right, if you have never modified the `busybox.config` then it goes to the default one. 

I'd do this:

```
cp build/busybox-1.27.2/.config ../src/<full system name>/busybox.config
```

Then run `make menuconfig` to set `BR2_PACKAGE_BUSYBOX_CONFIG` to `${NERVES_DEFCONFIG_DIR}/busybox.config`, and then `make savedefconfig`.

Then when I wanted to update it in the future, `make busybox-update-config` will work.

Side note: Busybox only knows how to save `.config` files. Linux and Buildroot can save `.config` and `_defconfig` versions where the `defconfig` version only has the nondefault settings.

Damirados:
I think at this point it is improvement over previous version. Would be nice for someone to check it out and test it, so it's not only me.

guides/advanced/building-sytems.md
```diff
@@ -277,34 +287,77 @@ Customizing your Nerves system is an advanced but powerful way to tailor the sys
 
 Customizing the build allows you to tailor the Nerves system to meet specific requirements for your hardware or application. This involves modifying Buildroot configurations and applying changes to the Nerves system.
 
-### Modify Buildroot Configuration
+### Modify Buildroot Package Configuration
 
-Nerves systems use Buildroot for building firmware. The `make menuconfig` command opens a menu-based interface where you can modify the Buildroot configuration:
+Navigate to the output directory of system you wish to modify.
```
mlainez, line 292:
```suggestion
Navigate to the output directory of the system you wish to modify.
```

guides/advanced/building-sytems.md
```diff
@@ -277,34 +287,77 @@ Customizing your Nerves system is an advanced but powerful way to tailor the sys
...
 
 ```bash
 cd o/<system short name>
-make menuconfig
 ```
 
-In this interface, you can:
+Workflow for customizing a Nerves system is the same as when using Buildroot outside of Nerves,
```
mlainez, line 298:
```suggestion
The workflow for customizing a Nerves system is the same as when using Buildroot outside of Nerves,
```

guides/advanced/building-sytems.md
```diff
@@ -277,34 +287,77 @@ Customizing your Nerves system is an advanced but powerful way to tailor the sys
...
+    ```
+
+    Like the Linux configuration, the Buildroot configuration will need to be
+    updated to point to the custom config. This can be done via `make menuconfig`
+    and navigating to **Target Packages** and finding the
+    **BusyBox configuration file to use** option under the
+    **BusyBox** package.
+
+    > NOTE: If your system uses **Additional BusyBox configuration fragment files**
+    option under it needs to be disabled for `make busybox-update-config` to work.
```
mlainez, line 345: Really ? I didn't know. I'll try it.

mlainez (commented):
Some additional comments/suggestions

fhunleth:
@Damirados Thank you so much for writing this. It looks really good. Also, @mlainez, thanks for the review.

PR #1034, opened 2024-12-20
Added RPI5 entry in supported-targets.md

peterHoburg:
Added RPI5 as a supported target. 
https://github.com/nerves-project/nerves_system_rpi5


fhunleth:
Thanks!

PR #1039, opened 2025-01-06
Create Contributing guide information

lawik:
Capturing automated checks that people can expect as well as actually trying to capture conventions and practices that are expected.

### DIFF
--- a/guides/core/contributing.md
+++ b/guides/core/contributing.md
@@ -1,0 +1,30 @@
+# Contributing
+
+We welcome contributions to Nerves. Issues, Pull Requests, knowledge sharing and questions
+in the various venues are all welcome. When contributing code there are some practicalities
+that are helpful to make the process smooth.
+
+Nerves and the various repos related to it have a common coding standard that is intended to
+apply everywhere. There are some automated ones and then there is some preferred practices
+that are currently manually encouraged.
+
+## Automated checks
+
+The tedious bits are mostly manageable through automated tools. If you want
+to avoid fixing CI failures you can run them locally first:
+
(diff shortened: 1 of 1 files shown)


lawik:
@fhunleth I know you have more and you might have opinions on what I included. Just edit away. I wanted to get this started but I don't think I can finish it without input. And I suppose there are just a few items and a couple of minutes to bang this out. Then we can look at making Credo plugins or something :)

fhunleth:
@lawik I think this is great, but your PR reminded me that we already have a contributing guide. I hadn't looked at it in a really long time and feel that it's not only out of date now, but also missing some new things that you added. Here's a link to the old guide. https://github.com/nerves-project/.github/blob/main/CONTRIBUTING.md. Let me know if you think it's best to refresh it or do something else.

lawik:
This document could be changed to be a coding-standards one. That Contributing guide is kind of wide and isn't the best thing to link someone who is in the midst of contributing code. The other one may also need some review but seems good overall :)

It mentions a Nerves Radar, is that a thing and is it relevant for outside contributors to know?

fhunleth:
> It mentions a Nerves Radar, is that a thing and is it relevant for outside contributors to know?

Oh gosh. That needs to be deleted.

I was under the impression that the `CONTRIBUTING.md` guide would appear in the GitHub UI somewhere, but it doesn't seem to. Maybe we delete the whole thing and go with yours. I would like it to be discoverable on all Nerves-related projects. Perhaps we should move the discussion to Discord since you uncovered something that's going to hit every repository.

PR #1042, opened 2025-01-08
Add hook for post-processing filesystem

lawik:
Providing the config for :nerves, :firmware and the key `post_processing_script` will now allow `mix firmware` to pass a flag to rel2fw.sh that tells it what script to run when the filesystem is finalized.

This is the hook required to do filesystem signing for dm-verity and any other similar schemes.

Added some info about it to the advanced configuration guide. Ideally expanded to be a signing guide later when that has been more fully built out.

### DIFF
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -134,10 +134,21 @@ defmodule Mix.Tasks.Firmware do
         fwup_conf -> ["-c", Path.join(File.cwd!(), fwup_conf)]
       end
 
+    post_processing_script =
+      case firmware_config[:post_processing_script] do
+        nil -> []
+        script when is_binary(script) -> ["-s", script]
+      end
+
     fw = ["-f", fw_out]
     release_path = Path.join(Mix.Project.build_path(), "rel/#{otp_app}")
     output = [release_path]
-    args = args ++ fwup_conf ++ rootfs_overlays ++ fw ++ rootfs_priorities ++ output
+
+    args =
+      args ++
+        fwup_conf ++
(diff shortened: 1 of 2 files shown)


lawik:
Requires this: https://github.com/nerves-project/nerves_system_br/pull/832

lawik:
CI hit a cyclomatic complexity limit. I am not convinced the code would be better by breaking something out into a separate function. It works pretty much in line with the rest. It is certainly possible to address it but it feels like it would take some kind of odd work?

Happy for input.

lawik:
Okay, found a decent way of reducing the cyclomatic complexity while retaining the clarity of what piece of config one is pulling in.

fhunleth:
Cool. I was hoping @jjcarstens would be able to get some time to review. He's been busy, so if he doesn't pop up, I'll make a pass and get it in a release.

lib/mix/tasks/firmware.ex
```diff
@@ -296,4 +288,31 @@ defmodule Mix.Tasks.Firmware do
       """)
     end
   end
+
+  defp config_arg(:rootfs_overlay, firmware_config) do
```
jjcarstens, line 292:
```suggestion
  defp rootfs_overlay_arg(firmware_config) do
```

LGTM! Though I would probably suggest descriptive purpose built functions for these args rather than a generic `config_arg` that has variable key arg

lawik: I liked the single function where all the config keys were exposed to find. But since I had to reduce complexity for the cyclomat I wanted to do something that made it visually simpler but kept the keys visible. Okay to do the suggested change but it becomes a bit of a hunt to figure out what's needed or involved when diving in I think.

jjcarstens (approved):

PR #1047, opened 2025-01-29
Add platform using macro back to fix nerves_toolchain_ctng error

fhunleth:
This is a partial revert of e0e249d1fad11c7b02e4f2f8d15951da9edd0d2b.


lib/nerves/package/platform.ex
```diff
@@ -66,4 +66,15 @@ defmodule Nerves.Package.Platform do
   artifact to the local build_path location.
   """
   @callback build_path_link(package :: Nerves.Package.t()) :: build_path_link :: String.t()
+
+  # Deprecated. Avoid using in new code.
```
jjcarstens, line 70: Does `@deprecated` work for macros? Is it worth having a little warning?

jjcarstens (approved):

PR #1092, opened 2025-10-16
Allow :nerves, :firmware, :fwup_conf option to be an absolute path

lawik:


lib/mix/tasks/firmware.ex
```diff
@@ -314,8 +314,25 @@ defmodule Mix.Tasks.Firmware do
...
     case firmware_config[:fwup_conf] do
-      nil -> []
-      fwup_conf -> ["-c", Path.join(File.cwd!(), fwup_conf)]
+      nil ->
+        []
+
+      fwup_conf ->
+        path =
+          cond do
+            File.exists?(fwup_conf) ->
```
fhunleth: It seems like `Path.expand/1` should be used here since it handles both absolute and relative paths.

lib/mix/tasks/firmware.ex
```diff
@@ -314,8 +314,19 @@ defmodule Mix.Tasks.Firmware do
...
-      fwup_conf -> ["-c", Path.join(File.cwd!(), fwup_conf)]
+      nil ->
+        []
+
+      fwup_conf ->
+        path = Path.expand(fwup_conf)
+
+        if not File.exists?(path) do
+          Mix.raise(
+            "File referenced in :nerves, :firmware, :fwup_conf not found. Tried: #{fwup_conf} and #{Path.join(File.cwd!(), fwup_conf)}"
```
fhunleth:
```suggestion
            "File referenced in :nerves, :firmware, :fwup_conf not found. Tried: #{Path.expand(fwup_conf)}"
```

PR #1109, opened 2026-01-12
Add mix task for discovering Nerves devices

fhunleth:
This works by using the operating system's mDNS stack or a generic
OTP-only one for querying the network. The OS stack is preferred since
it can report back cached entries more quickly and recent versions of
macOS prevent user applications from using mDNS directly.

Here's an example run:

```
$ mix nerves.discover
Discovering Nerves devices (waiting up to 5000ms)...

NAME         IP             SERIAL            VERSION  PRODUCT     PLATFORM  UUID
nerves-0316  192.168.7.128  55e77bfdd5030316  0.2.1    kiosk_demo  rpi5      cbf8c271-2673-51f7-ab08-3de09af404eb
nerves-8465  192.168.7.48
```

The first device advertises firmware attributes which is currently an
opt-in feature. The second device is found via a heuristic that looks
for the ssh service on devices with hostnames that start with "nerves".

Most logic is in the `nerves_discovery` library so that it can be used
in other programs too.

Currently not all fields are shown in the table since the rows can get
really long. The UUID get elided in the current implementation when the
terminal width is too small.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# SPDX-FileCopyrightText: None
+# SPDX-License-Identifier: CC0-1.0
+
+# Default PR assignments to the core-team
+* @nerves-project/core-team


lawik (approved):
See question, totally works on my linux machine and on my MacOS machine across Tailscale apparently :)

PR #1121, opened 2026-02-02
nerves.loadpaths: Do not crash when invoked without args

bjyoungblood:
This prevents a crash when ElixirLS is building a Nerves project with
MIX_TARGET set to something other than host.


fhunleth:
Thanks!

PR #1128, opened 2026-03-01
Support other env vars when determining if download progress should be shown

joshk:
GitHub Actions and CircleCI both set `CI` env to `true`.

`DEBIAN_FRONTEND=noninteractive` is commonly used to convey that interactive input isn't supported.

These would allow CI output to be cleaner without having to set `NERVES_LOG_DISABLE_PROGRESS_BAR`


lawik (approved):
I would probably reverse the logic to chevk for the exact disabling value on CI and NERVES_DISABLE.. so == true.

But doesn't really matter.

PR #1148, opened 2026-04-05
Use OTP's built-in castore (since OTP 15)

fhunleth:
This removes the `:castore` dependency.


c4lliope:
[doc] https://www.erlang.org/doc/apps/public_key/public_key_app.html

PR #1153, opened 2026-04-07
Recommend fwup_delta for creating firmware patches

fhunleth:
This shows a warning if trying to use `mix firmware.patch`. The docs
also have been updated to talk more generically about patches and leave
generation details to `fwup_delta`. Currently `fwup_delta` is so much
better than `mix firmware.patch` that anyone using it should switch.


fhunleth:
@lawik Please review. I felt bad removing the example, but that much of this information will be easier to maintain in `fwup_delta` and the most important thing that can be done here is to point people to that project.

lawik (approved):
Looks good.

PR #1157, opened 2026-04-15
Simplify resolvers and switch to explicit 2-phase expansion

fhunleth:
Resolvers used to be expanded and then run to make their downloads. This
moves a little more work to the expansion phase (aka the planning phase)
to lock everything down. The expansion previously resulted in a lot of
temporary values moving around only to be changed or ignored later.
Everything is explicit now. If you are debugging and need to see the
full list of URLs that will be tried, that's easy to get by just dumping
the plan list.

While doing this I saw that the GitHub download methods, `git_api` and
`git_release` evolved to be almost identical. The main difference was
whether to send the auth token, and even that was made moot since
sending the auth token was needed to get past API rate limiting. On top
of that, the API didn't need to be called. Perhaps early on it needed
this, but not any more.

The unit tests needed many updates. In order to make this easier to
review, I tried to keep the test names. There's a future step makes the
unit tests better organized and deletes redundant tests.

### DIFF
--- a/lib/nerves/artifact.ex
+++ b/lib/nerves/artifact.ex
@@ -15,6 +15,8 @@ defmodule Nerves.Artifact do
   alias Nerves.Artifact.Cache
   alias Nerves.Artifact.Resolvers
 
+  @resolver_modules [Resolvers.GithubAPI, Resolvers.GiteaAPI, Resolvers.URI]
+
   @checksum_short 7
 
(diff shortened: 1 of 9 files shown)


fhunleth:
I set up a private GitHub repo and this doesn't work. I think I got some bad information, so going to verify...

PR #1163, opened 2026-04-22
Updated the arch install docs

nomtrosk:
I had a fresh Arch system and these basic packages seem to be missing. Also the fwup instructions was missing.. Maybe they should have their own section?


fhunleth:
Thanks for updating this! I think we can take a pass on centralizing the fwup instructions in the future and it certainly isn't hurting anything here.

PR #1164, opened 2026-04-30
Add `:extra_rootfs_overlays` config for tools to inject overlays

jimsynz:
## Summary

- Adds `:extra_rootfs_overlays` to `config :nerves, :firmware` — a list of overlay paths concatenated with `:rootfs_overlay` before being passed to `rel2fw.sh -a`.
- Lets tools that distribute pre-built rootfs overlays append paths programmatically without overwriting the user's own `:rootfs_overlay` setting.
- Documented in the firmware task's moduledoc; `CHANGELOG.md` entry under `(unreleased)`.

## Motivation

While prototyping a curated Hex repository for distributing Buildroot-built target packages as Hex artefacts (NBPR), I needed a Mix task running before `mix firmware` to inject extracted package overlays into the rootfs. The existing `:rootfs_overlay` is the user's own; clobbering it from a tool isn't acceptable, and reading-then-rewriting it programmatically is racy across multiple tools that might want to do the same thing.

An additive key avoids both problems and is generally useful for any tool that wants to layer onto the firmware build (CI helpers that bake in build IDs, vendor SDK injection, etc.) — not just NBPR.

## Implementation

Two small changes in `lib/mix/tasks/firmware.ex`:

1. `build_firmware/3` now reads both `:rootfs_overlay` and `:extra_rootfs_overlays` and concatenates them before the existing `prevent_overlay_overwrites!` and `-a` arg assembly.
2. New `config_arg(:extra_rootfs_overlays, _)` clause that returns `[]` when unset and the user-provided list otherwise (no `Path.expand`, matching the list-case behaviour of `:rootfs_overlay`).

## Test plan

- [x] Existing `mix test` suite passes (142 tests, 0 failures), `mix format --check-formatted` clean
- [x] End-to-end against a real Nerves project: a Mix task sets `:extra_rootfs_overlays`, `mix firmware` honours it, the fetched overlay's contents land in the built `.fw` rootfs
- [ ] Maintainer review for naming + doc placement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@
 
 # Changelog
 
+## (unreleased)
+
+* New features
+  * Added `:extra_rootfs_overlays` option to `config :nerves, :firmware`. Tools
+    that need to inject rootfs overlays programmatically (e.g. package
+    distribution libraries) can now append paths without overwriting the
+    user's own `:rootfs_overlay` setting. The two lists are concatenated
+    before being passed to `rel2fw.sh`.
+
 ## v1.14.1 - 2026-04-22
 
 This release makes `mix nerves.system.shell` work again. We still recommend
--- a/lib/mix/tasks/firmware.ex
+++ b/lib/mix/tasks/firmware.ex
@@ -23,6 +23,17 @@ defmodule Mix.Tasks.Firmware do
     * `--verbose` - produce detailed output about release assembly
     * `--output` - (Optional) The path to the .fw file used to write the patch
       firmware. Defaults to `Nerves.Env.firmware_path/1`
+
+  ## Configuration
+
+  Set under `config :nerves, :firmware` in your project's config:
+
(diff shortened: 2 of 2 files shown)


fhunleth:
@jimsynz Could you give more background on the issue? The reason is that I've rejected proposals for this in the past since the convention is to install into the `priv` directory. That keeps all of the Mix release semantics in tact and you don't have to worry about hex packages stomping on each other either.

fhunleth:
Actually, I skimmed your issue too quickly. Sorry - I've gotten into the habit of skimming Claude-generated PRs due to their verbosity and yours actually wasn't all that bad. I totally understand your motivation. I still really like everything being under priv, but maybe there's a way to phase that in.

jimsynz:
Thanks for the direction @fhunleth - I've changed the approach. The only time I can think of that we actually need to modify the rootfs is in the cast of out-of-tree kernel modules (eg `zfs`) but we can work around that my copying the files directly into the rootfs before the image is created. Closing.

PR #1174, opened 2026-05-22
Raise loudly when NERVES_SYSTEM or NERVES_TOOLCHAIN is set to a non-directory path

kybishop:
This resolves an issue I've run into multiple times where Nerves silently fails to pick up your intended NERVES_SYSTEM, leading to many hours debugging to figure out I simply hadn't extracted the compressed package 🤦 

Previously, if NERVES_SYSTEM (or any Nerves artifact env var) was set to a non-empty value that wasn't a directory — for example, a .tar.gz artifact path — Artifact.dir/1 silently fell back to the default computed artifact path. The build would complete successfully using the wrong system, with no indication the override was ignored.

The fix is in Artifact.dir/1, which is the single decision point for "use env override or computed default." It now distinguishes three states: set and a valid directory (use it), set but not a directory (raise), and unset/empty (fall back to default). The error message names the variable, shows the bad path, and hints at the tar.gz case. This affects all callers of Artifact.dir/1:

- Nerves.Env.system_path/0 — called from bootstrap/0 during mix firmware and related tasks
- Nerves.Env.toolchain_path/0 — called from bootstrap/0 for the same
- Mix.Tasks.Nerves.Artifact.Details — called when running mix nerves.artifact.details <package>

### DIFF
--- a/lib/nerves/artifact.ex
+++ b/lib/nerves/artifact.ex
@@ -196,11 +196,24 @@ defmodule Nerves.Artifact do
   """
   @spec dir(Nerves.Package.t()) :: String.t()
   def dir(pkg) do
-    if env_var?(pkg) do
-      System.get_env(env_var(pkg)) |> Path.expand()
-    else
-      base_dir()
-      |> Path.join(name(pkg))
+    var_name = env_var(pkg)
+    var_value = System.get_env(var_name)
+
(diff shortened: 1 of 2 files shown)


fhunleth:
Makes sense. Thanks!

PR #1190, opened 2026-07-10
Add an Apple `container` build runner for macOS (Apple Silicon)

tomfarm:
Hey!

I build a few custom Nerves systems on macOS, and for a while I was switching
between local Docker, a remote Linux box and a local UTM Linux VM as a build
server — and I was never really happy with it (I'm honestly not so much of a
Docker fan). But the M-series Macs are so fast :) so I really wanted to build
directly on the Mac. So I sat down and looked into how the new Apple `container`
VMs work together with Nerves. First I did it only as a standalone library so I
could use it every day (https://github.com/chromasurf/nerves_container), and now
I would like to also offer it to core Nerves.

It adds a build runner `Nerves.Artifact.BuildRunners.Container` that runs the
Buildroot build in a small Apple `container` Linux VM instead of Docker. On Apple
Silicon with the `container` CLI installed it gets auto-selected for
`type: :system`, otherwise it falls back to Docker, and on Linux everything stays
on `Local` — so nothing changes for you unless you are on an Apple Silicon Mac
with `container` installed, and then it just works without extra config.

A few things I ran into and handled along the way:

- The build directory lives in a named EXT4 volume instead of a virtiofs bind
  mount — virtiofs is slow for the big Buildroot tree and can't even create the
  symlinks from the extracted tarball (apple/container#1209). The download cache
  and the sources are still bind-mounted from the host.
- Buildroot's ccache is kept on the host so it survives between builds. This was
  the biggest one for me — it makes rebuilding WebKit-heavy systems fast again.
- Memory-hungry builds (WebKit again 😅) were getting their compilers OOM-killed,
  so the guest gets a swapfile and can spill to disk like a normal Linux build VM.
- Plus a few env knobs for CPUs/memory/swap, and a small macOS 15 networking
  quirk that I documented with a one-line fix.

I've been running this on our own systems (three CM4 boards, two of
them with WebKit/Cog). For a normal rpi4-style system it works really well; in a
quick fair benchmark it was even a little bit faster than Docker and my UTM VM.

Two caveats: I only tested it on macOS 15 (Sequoia), I could not try
macOS 26 yet where Apple reworked a lot of the container networking, so I am not
sure how it behaves there. And I fully understand if this does not fit into core
Nerves — it adds one more "which runner builds what, and where" dimension and
that can get confusing quickly. If you prefer not to carry it I will keep the standalone package published so anyone can use it in the meantime (or for good).

Either way I thought it is worth sharing. Thanks for Nerves, it is really great :)


fhunleth:
Hi @tomfarm! Thanks for this PR! I had been curious about Apple container because a few people had told me about really good experiences with it. I'm running macOS 26, so I can try that out.

This might take me a little bit of time to merge since I have a few things going on right now, but rest assured, you sent this at a really good time for me to incorporate in Nerves.

fhunleth (approved):
This works for me on macOS 26. Code looks good. Thank you!

